PR #6, opened 2020-04-21
Update changelog ID to avoid collision with previous versions

boubaker:
Previously, before version 2.0.0-M25, the Changelog IDs 18 and 19 was used, and then deleted in version 2.0.0-M25 (Bad practice).
Thus with databases that had deployed 2.0.0-M25 version will not accept again those changelog IDs.
Thus, here we will increment it to 20.


mkrout (approved):
Lgtm

PR #22, opened 2020-05-14
Change popular spaces portlet permission temporarily to avoid security issue with displaying hidden/closed spaces

azayati:


rdenarie (approved):
LGTM

PR #26, opened 2020-06-12
Some labels are missing in dw homepage on the gamification gadget

rdenarie:
To address this issue, we remove the gamification properties without suffixe, because the default file is Gamification_en.properties.


thomasdelhomenie:
Shouldn't you also add back the strings removed in https://github.com/Meeds-io/gamification/commit/dd1936b307360d94c77748e52fe8b00c9205f9ba#diff-1b1cb9ad0f221960ddcb71ff9cfcb638L300-L306 ?

rdenarie:
This label is present in Gamification_en.properties : 
https://github.com/Meeds-io/gamification/blob/develop/services/src/main/resources/locale/addon/Gamification_en.properties#L20

thomasdelhomenie:
I don't see "homepage.profileStatus.header" and the following ones, is it expected ?

rdenarie:
Good catch. I removed the file but did not readd the missing labels.

thomasdelhomenie (approved):

PR #32, opened 2020-06-30
Update translations file

rdenarie:


boubaker (approved):

rdenarie:
We let this PR in waiting state : the rename may invalidate all translations. So we will do this later (not the day of the EOD)

boubaker (approved):

PR #41, opened 2020-07-23
Maint -27022: Some label not translated in the gamification rules management list

Rdidmohamedelhadi:
Problem : Some label not translated in the gamification rules management list so the fix is i added this label


ahamdi (approved):
LGTM

PR #58, opened 2020-09-08
MAINT-24816: Fix display of connections drawer.

mdiha:
Align the display of the connections drawer to exo drawers.


ahamdi (approved):
LGTM

PR #67, opened 2020-09-15
MAINT-27950: Make sure to display only user's first name in other users profile's welcome back portlet.

mdiha:
I made sure to check if it is the current user's profile or not in order to display the welcome back portlet accordingly.


ahamdi (approved):
LGTM

PR #68, opened 2020-09-15
Maint-28317 : Make sure to have the spaces of user profile

ahmadmido:
This fix assures to have only the user profile spaces in the drawer.


ahamdi (commented):
The PR is OK, 
but what will happen if I click on the number of the user's spaces ? cc @mzorai

mzorai:
> The PR is OK,
> but what will happen if I click on the number of the user's spaces ? cc @mzorai

When clicking on the number of spaces of the visited profile, a drawer should open to display commons spaces and visible spaces allowing me to join/request to join

ahamdi (approved):
LGTM
@mzorai let's open two other tasks for adding drawers for the common spaces & common connections that should be displayed when clicking on connections / spaces when we visit the profile of another user.

PR #87, opened 2020-12-02
Task-35609: Don't display external profile stats and badges overview blocks for an internal in external user profile page

azayati:


AzmiTouil (approved):
LGTM

PR #113, opened 2021-05-05
Override eslint rule We add a rule at upper level to prevent the use …

rdenarie:
…of axios.

The upper rule have error level, and fails the build
As we still use axios in gamification, we override the rule, to change the level to warn, waiting for change
When change applied in ecms, we will able to remove this override


boubaker (approved):

azayati (approved):
LGTM

PR #151, opened 2021-11-01
MAINT-47171: Fix the default gamification badge i18n description

hakermi:
**ISSUE**: The default description of the badge is always the default one even of the availability of its translation
**FIX**: Display the translation of the description while it's available before edit the description by the user


AzmiTouil (approved):
LGTM

PR #153, opened 2021-11-02
MAINT-47492: Fix gamification displaying badges by domain

hakermi:
**ISSUE**: A wrong passed argument of user global score when building user badges
**FIX**: Pass the right attribute of the global score to correctly counting the won badges of the user


AzmiTouil (approved):
LGTM

PR #156, opened 2021-11-10
MAINT-50896: Fix duplicated connections displays on connections drawer

hakermi:
**ISSUE**: Connections were pushed into an array each open of the drawer without being emptied of the old data
**FIX**: Empty the connections array on new data entries pushing to the array

### DIFF
--- a/portlets/src/main/frontend/src/apps/profileStats/components/ConnectionsDrawer.vue
+++ b/portlets/src/main/frontend/src/apps/profileStats/components/ConnectionsDrawer.vue
@@ -258,6 +258,7 @@ export default {
   methods: {
     init() {
       if (this.isCurrentUserProfile) {
+        this.connections = [];
         return this.getConnections(0);
       } else {
         return this.getConnections(0, 0).then(() => this.initPeopleSuggestionsList());


portlets/src/main/frontend/src/apps/profileStats/components/ConnectionsDrawer.vue
```diff
@@ -269,6 +269,7 @@ export default {
     },
     getConnections(offset, limit) {
       return getUserConnections('', offset, limit >= 0 ? limit : this.limit).then(data => {
+        this.connections.splice(0, this.connections.length);
```
AzmiTouil:
```suggestion
        this.connections = [];
```

AzmiTouil (changes requested):

portlets/src/main/frontend/src/apps/profileStats/components/ConnectionsDrawer.vue
ahamdi: You need to append new retrieved list to the already retrieved data. Is the offset updated on each click ? or are you loading all users on each click ?

hakermi: The getconnections() is called on each open of the drawer.

ahamdi: @hakermi if there is no load more , then the call will just return the first 20 connections. it is a bad UX since all Connections drawers let users paginate the list, couldn't we use an existing one instead of maintaining each one alone. WDYT @mzorai @AzmiTouil ?

hakermi: There is a load more but it's not paginated, you need to scroll to see the loaded entries

ahamdi: @hakermi could you fix the pagination loading then ?

mzorai: in fact @ahamdi the list of connections drawer was not paginated, it is a load more button + scroll to display the list
IMO we just fix the duplication issue in that task and create a new one (quick win) to add a pagination for the drawer (same as done for comments drawer in AS)
cc @hakermi

ahamdi: @mzorai Ok, I will validate this one

ahamdi (approved):

AzmiTouil (approved):

PR #165, opened 2021-11-26
Task-47492: Badges by domain is not displayed

AzmiTouil:
Prior this change,gamification points for normal activity in space are not earned, this is due to NPE when checking activity type


ahamdi:
@AzmiTouil I have a question, when the activity type is null ? when we post comments ?

AzmiTouil:
> @AzmiTouil I have a question, when the activity type is null ? when we post comments ?

When we post a normal activity (without attached file)

ahamdi:
@AzmiTouil then all textual space activities do have Null as activity type, which replaces the default activity type and uses the default activity template.

ahamdi (approved):

PR #170, opened 2021-12-06
TASK-51126 : User's profile: Spaces in common: Incorrect values

abenmbarek:
ISSUE : incorrect value of common spaces in user's profile and "see more spaces" button in the bottom of spaces drawer does not appear.
FIX : replace the value of "limitToFetch" which is never decalerd in the component with "-1" to get the correct size of common space list and display "see more spaces" button.

### DIFF
--- a/extension/pom.xml
+++ b/extension/pom.xml
@@ -22,7 +22,7 @@ Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
     <parent>
         <groupId>org.exoplatform.addons.gamification</groupId>
         <artifactId>gamification</artifactId>
-        <version>2.3.x-SNAPSHOT</version>
+        <version>2.3.x-maintenance-SNAPSHOT</version>
     </parent>
 
     <artifactId>gamification-extension</artifactId>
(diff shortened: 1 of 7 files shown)


rdenarie (changes requested):
It is not the good way to resolve the problem.
With your change, you load ALL spaces commons between the 2 users. If there is 500 spaces, you load all, even if you display the 10 first.

The rest service which get the common spaces return spaces you asked (from offset to offset+limit) and should return the total size of commons spaces. So, we must called firstly the rest service with offset=0 and limit=xxx (should be 10 but not sure), and then use the returnSize to display the common space number.
Then, when opening the drawer, we display the 10 first spaces (we already have it), and when clicks on show more, we make a second to get commons spaces from 10 to 20.

abenmbarek:
The rest service which get the common spaces does not return the correct size of the list from the backend when we use limitToFetch = 10, unlike the rest service that get the list of user spaces which gets the right size even when we use the limit condition, i'll check the backend of common spaces service.

rdenarie:
In fact, the returnSize should return always the total size, nevermind the limit, in our case the total size of common spaces between the 2 users.
So, if the rest service does not return the correct size, the fix must correct this behaviour to make the return size correct.

abenmbarek:
yep exactly, it's a rest service issue, i'm gonna fix the backend side.

PR #172, opened 2021-12-13
MAINT-52250: Fix missing gamification when post an activity

hakermi:
**ISSUE**: An **NPE** is thrown when checking the activity type in gamfication listener when the type of activity is null
**FIX**: Add a check of nullable value of the activity type to avoid the NPE

### DIFF
--- a/services/src/main/java/org/exoplatform/addons/gamification/listener/social/activity/GamificationActivityListener.java
+++ b/services/src/main/java/org/exoplatform/addons/gamification/listener/social/activity/GamificationActivityListener.java
@@ -292,8 +292,8 @@ public class GamificationActivityListener extends ActivityListenerPlugin {
   }
 
   private boolean isDocumentShareActivity(ExoSocialActivity activity) {
-    return activity.getType().equalsIgnoreCase("files:spaces") || activity.getType().equalsIgnoreCase("DOC_ACTIVITY")
-        || activity.getType().equalsIgnoreCase("contents:spaces");
+    return activity.getType() != null && (activity.getType().equalsIgnoreCase("files:spaces") || activity.getType().equalsIgnoreCase("DOC_ACTIVITY")
+            || activity.getType().equalsIgnoreCase("contents:spaces"));
   }
 
   private String getActivityUrl(ExoSocialActivity activity) {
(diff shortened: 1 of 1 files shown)


services/src/main/java/org/exoplatform/addons/gamification/listener/social/activity/GamificationActivityListener.java
```diff
@@ -292,6 +292,9 @@ private boolean isSpaceActivity(ExoSocialActivity activity) {
   }
 
   private boolean isDocumentShareActivity(ExoSocialActivity activity) {
+    if (activity.getType() == null) {
+      return false;
+    }
     return activity.getType().equalsIgnoreCase("files:spaces") || activity.getType().equalsIgnoreCase("DOC_ACTIVITY")
         || activity.getType().equalsIgnoreCase("contents:spaces");
```
rdenarie: Can you make all the test in the return ? without a if

rdenarie (changes requested):

services/src/main/java/org/exoplatform/addons/gamification/listener/social/activity/GamificationActivityListener.java
hakermi: done

services/src/main/java/org/exoplatform/addons/gamification/listener/social/activity/GamificationActivityListener.java
```diff
@@ -292,8 +292,8 @@ private boolean isSpaceActivity(ExoSocialActivity activity) {
   }
 
   private boolean isDocumentShareActivity(ExoSocialActivity activity) {
-    return activity.getType().equalsIgnoreCase("files:spaces") || activity.getType().equalsIgnoreCase("DOC_ACTIVITY")
-        || activity.getType().equalsIgnoreCase("contents:spaces");
+    return activity.getType() != null && (activity.getType().equalsIgnoreCase("files:spaces") || activity.getType().equalsIgnoreCase("DOC_ACTIVITY")
+            || activity.getType().equalsIgnoreCase("contents:spaces"));
```
rdenarie:
```suggestion
        || activity.getType().equalsIgnoreCase("contents:spaces"));
```

rdenarie (changes requested):

rdenarie (approved):

PR #185, opened 2021-12-28
Task 52257: wrong gamification points for like a comment

montassarhamadi:
Problem: Both  user1 and user2 received a gamification point  of Activity stream (Network): like comment.
Fix: the gamification points should be for User1:  Activity stream (space) like comment and User2:User2  Activity stream (space) receive like comment


services/src/main/java/org/exoplatform/addons/gamification/service/setting/rule/impl/RuleRegistryImpl.java
```diff
@@ -77,10 +77,9 @@ public void start() {
 
             for (RuleConfig rule : ruleMap.values()) {
                 RuleDTO ruleDTO = ruleService.findRuleByTitle(GAMIFICATION_DEFAULT_DATA_PREFIX+rule.getTitle());
-                if (ruleDTO == null) {
-                    store(rule);
+                if (ruleDTO == null || !(ruleDTO.getEvent().equals(rule.getEvent()))) {
```
ahamdi: why don't you check if the title is not the same ?

montassarhamadi: it's the  same but if i will check with title I must concatenate the title with this constant GAMIFICATION_DEFAULT_DATA_PREFIX so the condition will be :
if(ruleDTO == null || !(ruleDTO.getTitle().equals(GAMIFICATION_DEFAULT_DATA_PREFIX+rule.getTitle()))

ahamdi (approved):

PR #187, opened 2021-12-31
TASK-52175 : fix editing domain failure just after the add

abenmbarek:
ISSUE : add new domain and edit it without refreshing the list => nothing changed.

FIX : call the api to get the list of domains after the add action to refill the listdomains variable and get the right domain's id for the update action.

### DIFF
--- a/portlets/src/main/frontend/src/apps/components/domain/ManageDomains.vue
+++ b/portlets/src/main/frontend/src/apps/components/domain/ManageDomains.vue
@@ -96,7 +96,6 @@ export default {
   },
   methods: {
     getAllDomains(){
-    console.log('domains are displayed !');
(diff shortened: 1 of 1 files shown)


portlets/src/main/frontend/src/apps/components/domain/ManageDomains.vue
```diff
@@ -92,6 +92,11 @@ export default {
   },
   data: initialData,
   created() {
+    this.getAllDomains();
+  },
+  methods: {
+    getAllDomains(){
+    console.log('domains are displayed !');
```
hakermi:
```suggestion
```

rdenarie (approved):

PR #178, opened 2021-12-16
Task-52257: wrong gamification points for like a comment

montassarhamadi:
Problem:  Both  user1 and user2 received a gamification point  of Activity stream (Network): like comment.
Fix: the gamification points should be for User1:  Activity stream (space) like comment and for User2: Activity stream (space) receive like comment

PR #188, opened 2022-01-03
Task 52257: wrong gamification points for like a comment (#185)

eromdhaneexo:
Problem: Both  user1 and user2 received a gamification point  of Activity stream (Network): like comment
Fix: the gamification points should be : for User1:  Activity stream (space) like comment and for User2  Activity stream (space) receive like comment.


rdenarie (changes requested):
The fix contains 2 PR : 
- https://github.com/Meeds-io/gamification/pull/178
- https://github.com/Meeds-io/gamification/pull/185
But I'm afraid that you get only one (the second one).
Can you check ?
Thanks

rdenarie (approved):

PR #189, opened 2022-01-04
TASK-39450 : Deactivated rules still appear in gamification earn points page

abenmbarek:
ISSUE : get all the rules without filtering by enabled filed On How can I earn points page.

FIX : add new service to get the list of all enabled rules and use it On How can I earn points page + add the unit tests for the service, dao and the rest package ( i've noticed that there is no test class available for RulesRestTest so i added a fresh one ) + fix the name of a rule DTO mapper's method to be more appropriate with the english lang .

### DIFF
--- a/services/src/main/java/org/exoplatform/addons/gamification/entities/domain/configuration/RuleEntity.java
+++ b/services/src/main/java/org/exoplatform/addons/gamification/entities/domain/configuration/RuleEntity.java
@@ -31,10 +31,6 @@ import java.util.Objects;
                 query = "SELECT rule FROM Rule rule WHERE rule.isDeleted = false"
         ),
         @NamedQuery(
-                name = "Rule.getActiveRules",
-                query = "SELECT rule FROM Rule rule WHERE rule.isDeleted = false AND rule.isEnabled = true"
-        ),
-        @NamedQuery(
                 name = "Rule.getEnabledRules",
                 query = "SELECT rule FROM Rule rule where rule.isEnabled = :isEnabled AND rule.isDeleted = false"
         ),
--- a/services/src/main/java/org/exoplatform/addons/gamification/entities/domain/configuration/RuleEntity.java
+++ b/services/src/main/java/org/exoplatform/addons/gamification/entities/domain/configuration/RuleEntity.java
@@ -31,6 +31,10 @@ import java.util.Objects;
                 query = "SELECT rule FROM Rule rule WHERE rule.isDeleted = false"
         ),
         @NamedQuery(
+                name = "Rule.getActiveRules",
+                query = "SELECT rule FROM Rule rule WHERE rule.isDeleted = false AND rule.isEnabled = true"
+        ),
+        @NamedQuery(
                 name = "Rule.getEnabledRules",
                 query = "SELECT rule FROM Rule rule where rule.isEnabled = :isEnabled AND rule.isDeleted = false"
         ),
--- a/services/src/main/java/org/exoplatform/addons/gamification/rest/ManageRulesEndpoint.java
+++ b/services/src/main/java/org/exoplatform/addons/gamification/rest/ManageRulesEndpoint.java
@@ -103,14 +103,12 @@ public class ManageRulesEndpoint implements ResourceContainer {
 
         if (conversationState != null) {
             try {
-                List<RuleDTO> allRules = ruleService.getActiveRules();
+                List<RuleDTO> activesRules = ruleService.getActiveRules();
 
-                return Response.ok().cacheControl(cacheControl).entity(allRules).build();
+                return Response.ok().cacheControl(cacheControl).entity(activesRules).build();
 
             } catch (Exception e) {
-
                 LOG.error("Error listing active rules ", e);
-
                 return Response.serverError()
                         .cacheControl(cacheControl)
                         .entity("Error listing active rules")
@@ -123,8 +121,6 @@ public class ManageRulesEndpoint implements ResourceContainer {
                     .entity("Unauthorized user")
                     .build();
         }
-
-
     }
 
     @POST
--- a/services/src/main/java/org/exoplatform/addons/gamification/rest/ManageRulesEndpoint.java
+++ b/services/src/main/java/org/exoplatform/addons/gamification/rest/ManageRulesEndpoint.java
@@ -94,6 +94,39 @@ public class ManageRulesEndpoint implements ResourceContainer {
 
     }
 
+    @GET
+    @Path("/active")
+    @RolesAllowed("users")
+    public Response getActiveRules(@Context UriInfo uriInfo, @Context HttpServletRequest request) {
+
+        ConversationState conversationState = ConversationState.getCurrent();
+
+        if (conversationState != null) {
+            try {
+                List<RuleDTO> allRules = ruleService.getActiveRules();
+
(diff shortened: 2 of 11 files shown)


services/src/main/java/org/exoplatform/addons/gamification/entities/domain/configuration/RuleEntity.java
```diff
@@ -30,6 +30,10 @@
                 name = "Rule.getAllRules",
                 query = "SELECT rule FROM Rule rule WHERE rule.isDeleted = false"
         ),
+        @NamedQuery(
```
rdenarie: You should use the existing NamedQuery Rule.getEnabledRules
No need to add a new one

services/src/main/java/org/exoplatform/addons/gamification/rest/ManageRulesEndpoint.java
```diff
@@ -92,6 +92,39 @@ public Response getAllRules(@Context UriInfo uriInfo, @Context HttpServletReques
...
+    @GET
+    @Path("/active")
+    @RolesAllowed("users")
+    public Response getActiveRules(@Context UriInfo uriInfo, @Context HttpServletRequest request) {
+
+        ConversationState conversationState = ConversationState.getCurrent();
+
+        if (conversationState != null) {
+            try {
+                List<RuleDTO> allRules = ruleService.getActiveRules();
```
rdenarie:
```suggestion
                List<RuleDTO> activesRules = ruleService.getActiveRules();
```

services/src/main/java/org/exoplatform/addons/gamification/rest/ManageRulesEndpoint.java
```diff
@@ -92,6 +92,39 @@ public Response getAllRules(@Context UriInfo uriInfo, @Context HttpServletReques
...
+
+                return Response.ok().cacheControl(cacheControl).entity(allRules).build();
```
rdenarie:
```suggestion
                return Response.ok().cacheControl(cacheControl).entity(activesRules).build();
```

services/src/main/java/org/exoplatform/addons/gamification/rest/ManageRulesEndpoint.java
```diff
@@ -92,6 +92,39 @@ public Response getAllRules(@Context UriInfo uriInfo, @Context HttpServletReques
...
+
+            } catch (Exception e) {
+
```
rdenarie: Remove empty line
```suggestion
```

services/src/main/java/org/exoplatform/addons/gamification/rest/ManageRulesEndpoint.java
```diff
@@ -92,6 +92,39 @@ public Response getAllRules(@Context UriInfo uriInfo, @Context HttpServletReques
...
+                LOG.error("Error listing active rules ", e);
+
```
rdenarie:
```suggestion
```

services/src/main/java/org/exoplatform/addons/gamification/rest/ManageRulesEndpoint.java
```diff
@@ -92,6 +92,39 @@ public Response getAllRules(@Context UriInfo uriInfo, @Context HttpServletReques
...
+                return Response.serverError()
+                        .cacheControl(cacheControl)
+                        .entity("Error listing active rules")
+                        .build();
+            }
+
```
rdenarie, line 117:
```suggestion
```

services/src/main/java/org/exoplatform/addons/gamification/rest/ManageRulesEndpoint.java
```diff
@@ -92,6 +92,39 @@ public Response getAllRules(@Context UriInfo uriInfo, @Context HttpServletReques
...
+        } else {
+            return Response.status(Response.Status.UNAUTHORIZED)
+                    .cacheControl(cacheControl)
+                    .entity("Unauthorized user")
+                    .build();
+        }
+
+
```
rdenarie:
```suggestion
        }
```

services/src/main/java/org/exoplatform/addons/gamification/service/configuration/RuleService.java
```diff
@@ -184,6 +184,27 @@ public RuleDTO findRuleByEventAndDomain (String ruleTitle, String domain) {
...
     }
+
+    /**
+     * Get all Rules from DB
+     * @return RuleDTO list
+     */
+    @ExoTransactional
+    public List<RuleDTO> getActiveRules() {
+        try {
+            //--- load all Rules
```
rdenarie:
```suggestion
            //--- load actives Rules
```

services/src/main/java/org/exoplatform/addons/gamification/storage/dao/RuleDAO.java
```diff
@@ -95,6 +95,18 @@ public RuleEntity findRuleByEventAndDomain(String event, String domain) throws P
 
     }
 
+    public List<RuleEntity> getActiveRules() throws PersistenceException {
+
+        TypedQuery<RuleEntity> query = getEntityManager().createNamedQuery("Rule.getActiveRules", RuleEntity.class);
```
rdenarie: change to use the existing query

services/src/test/java/org/exoplatform/addons/gamification/test/rest/TestManageRulesEndpoint.java
```diff
@@ -0,0 +1,223 @@
+/*
+ * This file is part of the Meeds project (https://meeds.io/).
+ * Copyright (C) 2020 Meeds Association
```
rdenarie:
```suggestion
 * Copyright (C) 2022 Meeds Association
```

rdenarie (changes requested):

services/src/main/java/org/exoplatform/addons/gamification/storage/dao/RuleDAO.java
```diff
@@ -95,6 +95,18 @@ public RuleEntity findRuleByEventAndDomain(String event, String domain) throws P
...
+
+        try {
+            return query.getResultList();
+        } catch (NoResultException e) {
+            return null;
```
rdenarie: I think it is better to return an empty list in case of error, instead of null

rdenarie (changes requested):

services/src/main/java/org/exoplatform/addons/gamification/storage/dao/RuleDAO.java
wassimmhamdi: @rdenarie  IMO Returning null instead of empty list might result in better performance, because empty lists need time to be created and discarded, and they need space to be stored. To minimize this inconvenience, we should return shared immutable empty lists.

rdenarie: You can return Collections.emptyList()

In fact, with an empty list, you dont create a special case to manage at upper level. It is clearer to read and clearer to use.

It is a development good practice, advised by Sonar.
You can find more informations here :
https://stackoverflow.com/questions/36779498/why-is-it-considered-good-practice-to-return-an-empty-collection

rdenarie (approved):

PR #190, opened 2022-01-04
TASK-52175 : Modified domains,rules and badges with old titles are displayed again on the list after restart server

abenmbarek:
ISSUES : after modifying domains, rules or badges title and restarting the server, the same item before the modification will appear again in addition to the new modified item + after deleting a domain, the confirmation popup is still displayed + add new domain and edit it without refreshing the list => nothing changed.

FIX : while starting the server, the domains, badges and the rules will be added only in case of their tables in the database are empty so they can be added + adding some missing fields related to rules and the badges mapper class + prevent changing the createdDate field in case of update action + call the api to get the list of domains after the add action to refill the listdomains variable and get the right domain's id for the update action.


rdenarie (changes requested):
Can you resolv the 2 sonar code smells ?
Thanks

rdenarie (approved):

PR #192, opened 2022-01-07
TASK-39450 : Deactivated rules still appear in gamification earn points page

abenmbarek:
ISSUE : get all the rules without filtering by enabled filed On How can I earn points page.

FIX : add new service to get the list of all enabled rules and use it On How can I earn points page + add the unit tests for the service, dao and the rest package ( i've noticed that there is no test class available for RulesRestTest so i added a fresh one ) + fix the name of a rule DTO mapper's method to be more appropriate with the english lang .


services/src/main/java/org/exoplatform/addons/gamification/service/configuration/RuleService.java
```diff
@@ -184,6 +185,27 @@ public RuleDTO findRuleByEventAndDomain (String ruleTitle, String domain) {
...
+            if (rules != null) {
+                return ruleMapper.rulesToRuleDTOs(rules);
+            }else{
+                return Collections.emptyList();
+            }
+
+        } catch (Exception e) {
+            LOG.error("Error to find active rules",e);
+            throw(e);
+        }
```
rdenarie: As ruleDAO.getActiveRules does not return null, and as persistence exception is managed in the dao layer, i think we could simplify this function : 
```suggestion
        List<RuleEntity> rules =  ruleDAO.getActiveRules();
        return ruleMapper.rulesToRuleDTOs(rules);
```

services/src/main/java/org/exoplatform/addons/gamification/storage/dao/RuleDAO.java
```diff
@@ -95,6 +96,18 @@ public RuleEntity findRuleByEventAndDomain(String event, String domain) throws P
...
+        TypedQuery<RuleEntity> query = getEntityManager().createNamedQuery("Rule.getEnabledRules", RuleEntity.class)
+                .setParameter("isEnabled",true);
+        try {
+            return query.getResultList();
+        } catch (NoResultException e) {
+            return Collections.emptyList();
+        }
+
+    }
+
```
rdenarie:
```suggestion
    public List<RuleEntity> getActiveRules() {

        
        try {
            TypedQuery<RuleEntity> query = getEntityManager().createNamedQuery("Rule.getEnabledRules", RuleEntity.class)
                .setParameter("isEnabled",true);
            return query.getResultList();
        } catch (PersistenceExcetion e) {
            LOG.error("blabla to complete",e);
            return Collections.emptyList();
        } catch (NoResultException e) {
            LOG.error("blabla to complete",e);
            return Collections.emptyList();
        }

    }

```

rdenarie (changes requested):

services/src/main/java/org/exoplatform/addons/gamification/storage/dao/RuleDAO.java
abenmbarek: according to the compilation error, NoResultException should be catched the first and the only one, if i catch a second exception the compiler will ignore that catch scope.
**'catch' branch identical to 'NoResultException' branch** this warning message appears when use a second catch scope in this case.

rdenarie: did you try this syntax 
try {
     // stuff
} catch (Exception1 | Exception2 ex) {
     // Handle both exceptions
}

rdenarie (commented):
Can you fix the 2 code smells ?

abenmbarek:
already fixed but i don't know why sonarCloud did not recheck ?!

services/src/main/java/org/exoplatform/addons/gamification/service/configuration/RuleService.java
```diff
@@ -184,6 +184,17 @@ public RuleDTO findRuleByEventAndDomain (String ruleTitle, String domain) {
...
+
+    /**
+     * Get all Rules from DB
+     * @return RuleDTO list
+     */
+    @ExoTransactional
+    public List<RuleDTO> getActiveRules() {
+            //--- load actives Rules
+            List<RuleEntity> rules =  ruleDAO.getActiveRules();
+                return ruleMapper.rulesToRuleDTOs(rules);
```
rdenarie: Please check formatting

rdenarie (changes requested):

rdenarie (approved):

PR #193, opened 2022-01-11
TASK-52891: Disabled user is displayed in gamification Leaderboard.

IlhemEssaadi:
ISSUES : Disabled user is displayed in gamification Leaderboard.
FIX : the problem fixed by adding a condition to take only enabled users.


rdenarie (changes requested):
You cannot filter disable user after the request. 
Given a loadCapacity of 5. If the 5 first users provided by the query are disabled, your request will return nothing.

You should include the disabel criteria in leaderBoardFilter, and the query gamificationService.filter(leaderboardFilter); should return only enabled users

PR #194, opened 2022-01-13
TASK-53114:Add gamification connector to store challenges points

GouadriaHanen:
add new methods allow creating a new history with a specific rule


ahamdi (changes requested):
This should be closed as we changed the logic

PR #206, opened 2022-01-25
TASK-53450 : Fix NPE when we create Rule

GouadriaHanen:
CREATED_DATE should not be null when adding a new rule


Rdidmohamedelhadi (approved):

ahamdi (approved):

Rdidmohamedelhadi (commented):
Add description please !!

PR #229, opened 2022-02-09
TASK-53732: fixed consult winners list after click on show more button

GouadriaHanen:
while retrieving challenges after clicking on loadMore 
getAnnouncementsByChallenge takes the new offset and limit which may return a null list since the announcements number may not reach the offset

### DIFF
--- a/services/src/main/java/org/exoplatform/addons/gamification/rest/ChallengeRest.java
+++ b/services/src/main/java/org/exoplatform/addons/gamification/rest/ChallengeRest.java
@@ -180,8 +180,8 @@ public class ChallengeRest implements ResourceContainer {
       List<ChallengeRestEntity> challengeRestEntities = new ArrayList<>();
       for (Challenge challenge : challenges) {
         List<Announcement> challengeAnnouncements = announcementService.findAllAnnouncementByChallenge(challenge.getId(),
-                                                                                                       offset,
-                                                                                                       limit);
+                                                                                                       0,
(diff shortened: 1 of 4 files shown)


ahamdi:
@GouadriaHanen What is the required number of announcements to load by default ? cc @mzorai

mzorai:
@ahamdi already discussed with @GouadriaHanen and agreed to display 2 announcements by default

services/src/main/java/org/exoplatform/addons/gamification/rest/ChallengeRest.java
```diff
@@ -180,8 +180,8 @@ public Response getAllChallengesByUser(@ApiParam(value = "Offset of result", req
       List<ChallengeRestEntity> challengeRestEntities = new ArrayList<>();
       for (Challenge challenge : challenges) {
         List<Announcement> challengeAnnouncements = announcementService.findAllAnnouncementByChallenge(challenge.getId(),
-                                                                                                       offset,
-                                                                                                       limit);
+                                                                                                       0,
+                                                                                                       10);
```
ahamdi: @GouadriaHanen If we need just 2 then no need to put 10 as limit c c @mzorai

GouadriaHanen: done!

ahamdi (approved):

PR #231, opened 2022-02-10
TASK-XXXXX: Fix liquibase initialization for postgres

hbenali:


Rdidmohamedelhadi (changes requested):
Must not modify an already created changeSet. Another changeSet must be added !!

hbenali:
> Must not modify an already created changeSet. Another changeSet must be added !!

This fix will be applied on Fresh data instances, other instances won't be impacted by this change, since this changeset was tagged on `DATABASECHANGELOG` table managed by liquibase and has to be skipped ;)
Also `BIT` type is no more supported by Postgres which leads to errors and prevents liquibase from applying changesets

PR #235, opened 2022-02-11
TASK-53877 : fix displayingchallenge point and program

GouadriaHanen:
fix displaying challenge points and program in challenge card and challenge details
fix empty program after updating challenge

### DIFF
--- a/challenges-webapp/src/main/webapp/vue-app/challenges/components/ChallengeDrawer.vue
+++ b/challenges-webapp/src/main/webapp/vue-app/challenges/components/ChallengeDrawer.vue
@@ -255,7 +255,6 @@ export default {
       this.$refs.challengeDescription.disabled = false;
       this.warning= null;
       this.$refs.challengeProgram.program = null;
-      this.$refs.challengeProgram.broadcast = true;
     },
     open(){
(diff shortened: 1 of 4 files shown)


challenges-webapp/src/main/webapp/vue-app/challenges/components/ChallengeDrawer.vue
```diff
@@ -228,6 +228,7 @@ export default {
       }
       this.$refs.challengeProgram.broadcast = false;
       this.$refs.challengeProgram.program =  this.challenge.program;
+      this.$refs.challengeProgram.broadcast = true;
```
ahamdi, line 231: We can remove the line 229 instead of duplicating instruction

ahamdi (changes requested):

challenges-webapp/src/main/webapp/vue-app/challenges/components/ChallengeDrawer.vue
GouadriaHanen: it is not a duplication it is just enabling then disabling broadcast program change

ahamdi (approved):

PR #249, opened 2022-02-15
TASK-50956: add search filter to challenge app

GouadriaHanen:
add a search filter by title, description or program to challenge  app


challenges-webapp/src/main/webapp/vue-app/challenges/components/Challenges.vue
```diff
@@ -61,10 +70,19 @@ export default {
...
   }),
   computed: {
     classWelcomeMessage() {
       return !this.displayChallenges ? 'emptyChallenges': '';
+    },
+    challengesToDisplay() {
+      let challenges = this.challenges.slice();
+      if (this.search && this.search.trim().length) {
+        const searchTerm = this.search.trim().toLowerCase();
+        challenges = challenges.slice().filter(challenge => (challenge.title && challenge.title.toLowerCase().indexOf(searchTerm)) >= 0 || (challenge.description && challenge.description.toLowerCase().indexOf(searchTerm) >= 0) || (challenge.program && challenge.program.title && challenge.program.title.toLowerCase().indexOf(searchTerm) >= 0));
```
ahamdi, line 83: We need to load challenges from database and filter them, instead of just filtering challenges already loaded.

ahamdi (changes requested):

challenges-webapp/src/main/webapp/vue-app/challenges/components/Challenges.vue
GouadriaHanen: its just search by keyword filter to filter loaded challenges not an advanced filter

mzorai: It would be better to be a general filter, not the advanced one but a filter for ALL challenges not only the loaded ones

PR #258, opened 2022-02-23
TASK-53884:fix error message when access to challenge application with users non-admin

GouadriaHanen:
The list of domains is restricted for non-admins fixed by retrieving list of domains just for challenge  creators which are admins

### DIFF
--- a/challenges-webapp/src/main/webapp/vue-app/challenges/components/ProgramSuggester.vue
+++ b/challenges-webapp/src/main/webapp/vue-app/challenges/components/ProgramSuggester.vue
@@ -59,6 +59,12 @@
 
 <script>
 export default {
+  props: {
+    canAddChallenge: {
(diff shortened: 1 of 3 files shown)


challenges-webapp/src/main/webapp/vue-app/challenges/components/ProgramSuggester.vue
```diff
@@ -59,6 +59,12 @@
 
 <script>
 export default {
+  props: {
```
ahamdi, line 62: Where is this props filled with the value of canAddChallenge ?

GouadriaHanen: done!

ahamdi (approved):

PR #264, opened 2022-02-25
TASK-52803: Add permalink for challenges

GouadriaHanen:
- add a permalink to challenge details 
- error alert when accessing to challenge with permalink for user who is not member of the challenge's audience or not existing challenge

### DIFF
--- a/challenges-webapp/src/main/webapp/vue-app/challenges/components/Challenges.vue
+++ b/challenges-webapp/src/main/webapp/vue-app/challenges/components/Challenges.vue
@@ -46,6 +46,8 @@
         {{ $t('challenges.button.ShowMore') }}
       </v-btn>
     </v-row>
+    <challenge-details-drawer
+      ref="challengeDetails" />
   </v-app>
 </template>
 <script>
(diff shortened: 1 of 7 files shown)


challenges-webapp/src/main/webapp/vue-app/challenges/components/Challenges.vue
```diff
@@ -77,6 +79,22 @@ export default {
     });
     this.$root.$on('challenge-added', this.pushChallenge);
     this.$root.$on('challenge-updated', this.refreshChallenges);
+    const urlPath = document.location.pathname;
+    if (urlPath.includes('challenges/')) {
+      setTimeout(() => {
+        const challengeId = urlPath.split('challenges/')[1].split(/[^0-9]/)[0];
```
ahamdi: Could you check if this regex works ? could you try to find a regex that will capture the ID without the need to double split ?

services/src/main/java/org/exoplatform/addons/gamification/rest/ChallengeRest.java
```diff
@@ -100,8 +100,11 @@ public Response getChallengeById(@ApiParam(value = "Challenge technical id", req
     String currentUserId = Utils.getCurrentUser();
     try {
       Challenge challenge = challengeService.getChallengeById(challengeId, currentUserId);
-      List<Announcement> announcementList = announcementService.findAllAnnouncementByChallenge(challengeId, offset, limit);
-      return Response.ok(EntityMapper.fromChallenge(challenge, announcementList)).build();
+      if (challenge != null) {
```
ahamdi: I will be better if we do the opposite check and return a 404 error if the challenge is not found

ahamdi (changes requested):

ahamdi (approved):

PR #265, opened 2022-02-25
TASK-54694 : fix loading winners

GouadriaHanen:


services/src/main/java/org/exoplatform/addons/gamification/rest/AnnouncementRest.java
```diff
@@ -88,11 +100,24 @@ public Response getAllAnnouncementByChallenge(@ApiParam(value = "id of the chall
     if (limit <= 0) {
       return Response.status(Response.Status.BAD_REQUEST).entity("Limit must be positive").build();
     }
+    EntityTag eTag = null ;
+    List<AnnouncementRestEntity> announcementsRestEntities = new ArrayList<>();
+    eTag = new EntityTag(String.valueOf(announcementsRestEntities.hashCode()));
```
ahamdi: unneeded, you have to calculate Etag once you have loaded the entities to cache the response.

services/src/main/java/org/exoplatform/addons/gamification/rest/AnnouncementRest.java
```diff
@@ -24,10 +28,16 @@
...
   private static final Log    LOG = ExoLogger.getLogger(AnnouncementRest.class);
 
+  private final CacheControl cacheControl;
+
+
   private AnnouncementService announcementService;
 
   public AnnouncementRest(AnnouncementService announcementService) {
     this.announcementService = announcementService;
+    this.cacheControl = new CacheControl();
```
ahamdi: Are you using this ? why did you add it ?

ahamdi (changes requested):

services/src/main/java/org/exoplatform/addons/gamification/rest/AnnouncementRest.java
GouadriaHanen: done!

services/src/main/java/org/exoplatform/addons/gamification/rest/AnnouncementRest.java
GouadriaHanen: cleaned up!

ahamdi (approved):

PR #278, opened 2022-03-01
TASK-54796: fix rules duplication

GouadriaHanen:
- fixed duplicated rules after update
- added new resource bundle for some events

### DIFF
--- a/portlets/src/main/webapp/vue-app/rule/components/ManageRules.vue
+++ b/portlets/src/main/webapp/vue-app/rule/components/ManageRules.vue
@@ -157,10 +157,10 @@ export default {
     },
     updateRule(ruleDTO) {
       this.$RuleServices.updateRule(ruleDTO)
-        .then(rule => {
+        .then(() => {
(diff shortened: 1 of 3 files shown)


portlets/src/main/webapp/vue-app/rule/components/ManageRules.vue
```diff
@@ -157,10 +157,10 @@ export default {
     },
     updateRule(ruleDTO) {
       this.$RuleServices.updateRule(ruleDTO)
-        .then(rule => {
+        .then(() => {
           this.addSuccess = true;
           this.updateMessage = 'updated';
-          this.rules.push(rule);
+          this.rules =  this.getRules();
```
ahamdi: When updating we need to just replace the rule with the updated one and not reload all rules right ?

GouadriaHanen: done!

ahamdi (approved):

PR #282, opened 2022-03-04
TASK-54726 :Bad behaviors with dates 

GouadriaHanen:
fixes :
TASK-54726 : fix bad behaviors dates for challenges
TASK-54914: change ends icon color
TASK-54410: fix ability to create challenges with previous selected start date

### DIFF
--- a/challenges-webapp/src/main/webapp/vue-app/challenges/components/ChallengeDrawer.vue
+++ b/challenges-webapp/src/main/webapp/vue-app/challenges/components/ChallengeDrawer.vue
@@ -247,9 +247,6 @@ export default {
       this.challenge = {};
       this.$refs.challengeProgram.broadcast = true;
       this.$refs.challengeDatePicker.startDate = null;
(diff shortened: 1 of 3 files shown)


challenges-webapp/src/main/webapp/vue-app/challenges/components/ChallengeDrawer.vue
```diff
@@ -247,6 +247,9 @@ export default {
       this.challenge = {};
       this.$refs.challengeProgram.broadcast = true;
       this.$refs.challengeDatePicker.startDate = null;
+      if (this.$refs.challengeDatePicker && this.$refs.challengeDatePicker.$refs.challengeStartDate){
```
ahamdi: I have two concerns : 
 - Couldn't we just invoke a method in challengeDatePicker to reset that variable ? 
 - Shouldn't we do the same logic for the field endDate ?

GouadriaHanen: done!

ahamdi (approved):

PR #283, opened 2022-03-04
Task-54855: Fix issues with winners display

SaraBoutej:


challenges-webapp/src/main/webapp/vue-app/challenges/components/ChallengeCard.vue
```diff
@@ -169,8 +170,9 @@ export default {
         activityId: announcement.activityId,
         createDate: announcement.createdDate
       };
-      this.listWinners.unshift(newAnnouncement);
+      this.listWinners.push({'userName': newAnnouncement.user.remoteId});
       this.challenge.announcementsCount = this.challenge.announcementsCount +1;
+      this.$refs.announcementsUsersAvatar.refresh(newAnnouncement.user.remoteId);
```
AzmiTouil: Can we just send an event with the full user object to not re-retrieve in ExoUserAvatarsList.vue ?

AzmiTouil (changes requested):

challenges-webapp/src/main/webapp/vue-app/challenges/components/ChallengeCard.vue
SaraBoutej: done

AzmiTouil (approved):

PR #299, opened 2022-03-30
TASK-55602 Store Gamification System Domains on startup even if already populated first time

boubaker:
Prior to this change, when adding new Gamification Connectors with new domains, when the server has already a data populated into it, then the Domain isn't imported and consequently, the Domain isn't imported. This modification will attempt to import 'System' defined domains each startup when it doesn't exist.


rdenarie:
Does the task number in PR title is the good one ? Does this fix done in "move gamification connector" context ?

boubaker:
yes

mkrout (approved):
LGTM

PR #306, opened 2022-04-05
TASK-55998: fix dates issues

GouadriaHanen:
fixed issues by changing date.toInstant() by LocalDateTime when parsing in/To RFC3339Date since :
The LocalDateTime represents date-time without a time-zone such as 2019-10-25T12:15:30 whereas Instant is an instantaneous point on the time-line

### DIFF
--- a/services/src/main/java/org/exoplatform/addons/gamification/utils/Utils.java
+++ b/services/src/main/java/org/exoplatform/addons/gamification/utils/Utils.java
@@ -1,6 +1,8 @@
 package org.exoplatform.addons.gamification.utils;
 
 import java.sql.Timestamp;
+import java.time.LocalDateTime;
+import java.time.ZoneId;
 import java.time.ZoneOffset;
 import java.time.ZonedDateTime;
 import java.time.format.DateTimeFormatter;
@@ -121,7 +123,7 @@ public class Utils {
     if (StringUtils.isBlank(dateString)) {
       return null;
     }
-    ZonedDateTime zonedDateTime = ZonedDateTime.parse(dateString, RFC_3339_FORMATTER);
+    ZonedDateTime zonedDateTime = LocalDateTime.parse(dateString, RFC_3339_FORMATTER).atZone(ZoneId.systemDefault()).withZoneSameLocal(ZoneOffset.UTC);
     return new Date(Timestamp.valueOf(zonedDateTime.toLocalDateTime()).getTime());
   }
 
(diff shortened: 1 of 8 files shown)


services/src/main/java/org/exoplatform/addons/gamification/utils/Utils.java
```diff
@@ -114,16 +115,16 @@ public static String toRFC3339Date(Date dateTime) {
...
   }
 
   public static Date parseRFC3339Date(String dateString) {
     if (StringUtils.isBlank(dateString)) {
       return null;
     }
-    ZonedDateTime zonedDateTime = ZonedDateTime.parse(dateString, RFC_3339_FORMATTER);
-    return Date.from(zonedDateTime.toInstant());
+    ZonedDateTime zonedDateTime = LocalDateTime.parse(dateString, RFC_3339_FORMATTER).atZone(ZoneId.systemDefault()).withZoneSameLocal(ZoneOffset.UTC);
+    return new Date(Timestamp.valueOf(zonedDateTime.toLocalDateTime()).getTime());
```
boubaker:
```suggestion
    LocalDateTime localDateTime = LocalDateTime.parse(dateString, RFC_3339_FORMATTER);
    return Date.from(localDateTime.toInstant(ZoneOffset.UTC));
```

services/src/main/java/org/exoplatform/addons/gamification/utils/Utils.java
```diff
@@ -114,16 +115,16 @@ public static String toRFC3339Date(Date dateTime) {
     if (dateTime == null) {
       return null;
     }
-    ZonedDateTime zonedDateTime = ZonedDateTime.from(dateTime.toInstant().atOffset(ZoneOffset.UTC));
+    ZonedDateTime zonedDateTime = dateTime.toInstant().atZone(ZoneOffset.systemDefault()).withZoneSameLocal(ZoneOffset.UTC);
```
boubaker:
```suggestion
    ZonedDateTime zonedDateTime = dateTime.toInstant().atZone(ZoneOffset.UTC)
```

GouadriaHanen: done!

services/src/main/java/org/exoplatform/addons/gamification/utils/Utils.java
GouadriaHanen: done!

boubaker (approved):

PR #329, opened 2022-05-12
Task-56738: Could not create challenge. (#326)

montassarhamadi:
Problem: with SGBD PostgreSQL we cannot create challenge and an error is occured : "Error: Error saving challenge" and can't create challange.this is due a problem which is the autoIncrement of the column not functional CHALLENGE_MANAGER_ID in the table CHALLENGE_MANAGER_RULE for postgreSQL database.
Fix: added new constraint AutoIncrement on this column in the file changelog and now this fix works for both types of databases mysql and PostgreSQL.

### DIFF
--- a/services/src/main/resources/db/changelog/gamification.db.changelog-1.0.0.xml
+++ b/services/src/main/resources/db/changelog/gamification.db.changelog-1.0.0.xml
@@ -409,4 +409,14 @@ Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
             <column name="RULE_ID"/>
         </createIndex>
     </changeSet>
+    <changeSet id="1.0.0-35" author="exo-gamification">
+        <addAutoIncrement columnDataType="BIGINT" columnName="ID" incrementBy="1" startWith="1"
+                          tableName="GAMIFICATION_RULE"/>
+        <addAutoIncrement columnDataType="BIGINT" columnName="ID" incrementBy="1" startWith="1"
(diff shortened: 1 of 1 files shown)


services/src/main/resources/db/changelog/gamification.db.changelog-1.0.0.xml
```diff
@@ -409,4 +409,14 @@ Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
             <column name="RULE_ID"/>
         </createIndex>
     </changeSet>
+    <changeSet id="1.0.0-35" author="exo-gamification">
```
boubaker, line 412: Is this change tested on MySQL, it seems not. Could you check please

boubaker: BTW, why this change exists in develop and not stable/2.3.x ? it due to a commit that is not backported apparently https://github.com/Meeds-io/gamification/commit/978945a52862c3eea93216ecd5ae88a589d73539

montassarhamadi: > BTW, why this change exists in develop and not stable/2.3.x ? it due to a commit that is not backported apparently [978945a](https://github.com/Meeds-io/gamification/commit/978945a52862c3eea93216ecd5ae88a589d73539)

Yep, that's why I get a conflict when I tried to backport this fix. For MySQL SGDB, I tested it on maint-ft1 and I found it works

boubaker (approved):

PR #332, opened 2022-05-16
Task-54734: fix bulleted list indication in challenge details drawer

MayTekayaa:
Fix bulleted list indication when displaying description field  in chanllenge details drawer


GouadriaHanen (approved):
LGTM

PR #334, opened 2022-05-17
Task-57291: USe the display name instead of the userID in Realizations table

MayTekayaa:
In the realizations table, in Grantee and user fileds, the userID has been replaced with the display name in order to facilitate rewards count for the rewarding admin


GouadriaHanen (approved):
LGTM

PR #339, opened 2022-05-19
TASK-55183:fix displaying and recreating disabled domains

GouadriaHanen:


services/src/main/java/org/exoplatform/addons/gamification/service/setting/zone/impl/ZoneRegistryImpl.java
```diff
@@ -68,8 +60,8 @@ public void start() {
       try {
         // Processing registered domains
         for (ZoneConfig domain : zoneMap.values()) {
-          DomainDTO domainDTO = domainService.findDomainByTitle(domain.getZoneName());
-          if (domainDTO == null) {
+          DomainDTO domainDTO = domainService.getDomainByTitle(domain.getZoneName());
+          if (domainDTO == null || domainDTO.isDeleted()) {
```
ahamdi, line 64: If a domain is deleted, will it be recreated in next restart ?

services/src/test/java/org/exoplatform/addons/gamification/service/configuration/DomainServiceTest.java
```diff
@@ -67,10 +68,17 @@ public void testGetAllDomains() {
...
-    assertNotNull(domainService.findDomainByTitle(GAMIFICATION_DOMAIN));
+  public void testFindEnabledDomainByTitle() {
+    assertNull(domainService.findEnabledDomainByTitle(GAMIFICATION_DOMAIN));
+    DomainDTO domainDTO = newDomainDTO();
+    assertNotNull(domainService.findEnabledDomainByTitle(GAMIFICATION_DOMAIN));
+    domainDTO.setEnabled(false);
+    try {
+      domainService.updateDomain(domainDTO);
+    } catch (ObjectNotFoundException e) {
+    //
```
ahamdi, line 79: should it fail when it catches an exception ?

ahamdi: Could you explain when we use getDomainByTitle and when we use findEnabledDomainBytitle ?

services/src/main/java/org/exoplatform/addons/gamification/service/setting/zone/impl/ZoneRegistryImpl.java
GouadriaHanen: it is expected that the automatically created domains should be recreated if not found

services/src/test/java/org/exoplatform/addons/gamification/service/configuration/DomainServiceTest.java
GouadriaHanen: in our case for the achievement table after deactivating a domain the program ID field is empty, we need to display its program even if it is disabled we use in this case getDomainByTitle also for the zoneRegistry
 in other cases we use findEnabledDomainBytitle

services/src/main/java/org/exoplatform/addons/gamification/service/setting/zone/impl/ZoneRegistryImpl.java
ahamdi: @GouadriaHanen it will be better to forbid deleting them instead, WDYT? cc @mzorai

GouadriaHanen: there is too much behavior that should be forbidden (delete and rename) but as discussed with @mzorai  in the next iteration there will be too many changes in the behavior of the domains so we decided to just fix the bug in the achievement table and the double creating of disabled domain

mzorai: yes I'd confirm too many changes will be made on domains (will be named programs in next iteration) so IMO just fix the issue of recreating disabled domains when restart the server cc @GouadriaHanen @ahamdi

ahamdi (approved):

PR #348, opened 2022-05-25
TASK-53132: Enable posting announcement challenge with emoji 

ArfaouiMohamedAmine:
when an emoji is added to the description to announce a challenge, an sql error has been endorsed, it turns out that the description (named 'COMMENT' in the announecement table) is typed as varchar(2000), inorder to resolve this issue ; a type modification of utf8mb4 COLLATE utf8mb4_unicode_ci to support emojis has been added to the changelog. 
FYI: 
- locally the error is not reproduced .
- with this type modification it is still not reproduced for both mySQL qnd Postgres .

### DIFF
--- a/services/src/main/resources/db/changelog/gamification.db.changelog-1.0.0.xml
+++ b/services/src/main/resources/db/changelog/gamification.db.changelog-1.0.0.xml
@@ -419,7 +419,7 @@ Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
         <addAutoIncrement columnDataType="BIGINT" columnName="CHALLENGE_MANAGER_ID" incrementBy="1" startWith="1"
                           tableName="CHALLENGE_MANAGER_RULE"/>
     </changeSet>
-    <changeSet id="1.0.0-37" author="exo-gamification" dbms="postgresql">
+    <changeSet id="1.0.0-37" author="exo-gamification" dbms="mysql">
         <modifyDataType   columnName="COMMENT"
                           newDataType="NVARCHAR(2000) SET utf8mb4 COLLATE utf8mb4_unicode_ci"
                           tableName="GAMIFICATION_ACTIONS_HISTORY"/>
--- a/services/src/main/resources/db/changelog/gamification.db.changelog-1.0.0.xml
+++ b/services/src/main/resources/db/changelog/gamification.db.changelog-1.0.0.xml
@@ -419,7 +419,7 @@ Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
         <addAutoIncrement columnDataType="BIGINT" columnName="CHALLENGE_MANAGER_ID" incrementBy="1" startWith="1"
                           tableName="CHALLENGE_MANAGER_RULE"/>
     </changeSet>
-    <changeSet id="1.0.0-37" author="exo-gamification" dbms="mysql">
+    <changeSet id="1.0.0-37" author="exo-gamification" dbms="mysql,hsqldb,postgresql,h2,oracle,mssql">
         <modifyDataType   columnName="COMMENT"
                           newDataType="NVARCHAR(2000) SET utf8mb4 COLLATE utf8mb4_unicode_ci"
                           tableName="GAMIFICATION_ACTIONS_HISTORY"/>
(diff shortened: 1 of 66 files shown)


services/src/main/resources/db/changelog/gamification.db.changelog-1.0.0.xml
```diff
@@ -419,4 +419,10 @@ Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
         <addAutoIncrement columnDataType="BIGINT" columnName="CHALLENGE_MANAGER_ID" incrementBy="1" startWith="1"
                           tableName="CHALLENGE_MANAGER_RULE"/>
     </changeSet>
+    <changeSet id="1.0.0-37" author="exo-gamification" dbms="mysql">
```
GouadriaHanen, line 422: we need to change the type for all DB type not just mysql

GouadriaHanen (changes requested):

services/src/main/resources/db/changelog/gamification.db.changelog-1.0.0.xml
ArfaouiMohamedAmine: and by not specifying the dbms value, is it going to target all the db types by default ? should I delete the dbms or repeat this changeset for all the db types ?

GouadriaHanen: you need to specify the dbms value : ' mysql,mssql,h2,sybase,db2,hsqldb,postgresql" '

services/src/main/resources/db/changelog/gamification.db.changelog-1.0.0.xml
```diff
@@ -419,4 +419,10 @@ Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
         <addAutoIncrement columnDataType="BIGINT" columnName="CHALLENGE_MANAGER_ID" incrementBy="1" startWith="1"
                           tableName="CHALLENGE_MANAGER_RULE"/>
     </changeSet>
+    <changeSet id="1.0.0-37" author="exo-gamification" dbms="mysql,mssql,h2,sybase,db2,hsqldb,postgresql">
```
ahamdi: this changeset is specific to Mysql, do you reproduce the problem on Postgresql ?
Otherwise, we do support just Mysql and Postgresql, no need to mention other RDBMS

ahamdi (commented):
Please check my comment.

ahamdi (changes requested):
@ArfaouiMohamedAmine Please try to reproduce on local environment, you have to use the same collation for yopur local Mysql instance as the one used in Acceptance. 
We can not merge without knowing real cause. Could you take more time to investigate ?

services/src/main/resources/db/changelog/gamification.db.changelog-1.0.0.xml
ArfaouiMohamedAmine: roger that !
i'll verify Postgresql's influence on the problem 
I want to mention that locally the problem is not reproduced, it's reproduced on the acceptance .
the instance logs of acceptance are pointing out that it's an SQL column type issue 'COMMENT' column .

ahamdi:
@ArfaouiMohamedAmine Please update the description and the description with the title and description of what was done and why as explained in the PR best practices https://developer.exoplatform.org/#id-pull-requests-best-practices

ArfaouiMohamedAmine:
@ahamdi the requested modifications have been added .

ArfaouiMohamedAmine:
@ahamdi found out that locally the column type is not updated 
so this change we made will do the work .

services/src/main/resources/db/changelog/gamification.db.changelog-1.0.0.xml
```diff
@@ -419,4 +419,10 @@ Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
         <addAutoIncrement columnDataType="BIGINT" columnName="CHALLENGE_MANAGER_ID" incrementBy="1" startWith="1"
                           tableName="CHALLENGE_MANAGER_RULE"/>
     </changeSet>
+    <changeSet id="1.0.0-37" author="exo-gamification" dbms="mysql,mssql">
```
ahamdi:
```suggestion
    <changeSet id="1.0.0-37" author="exo-gamification" dbms="mysql">
```

ahamdi (approved):

ArfaouiMohamedAmine:
when an emoji is added to the description to announce a challenge, an sql error has been endorsed, it turns out that the description (named 'COMMENT' in the announecement table) is typed as varchar(2000), inorder to resolve this issue ; a type modification of utf8mb4 COLLATE utf8mb4_unicode_ci to support emojis has been added to the changelog.
FYI:

locally the error is not reproduced .
with this type modification it is still not reproduced for both mySQL .

PR #352, opened 2022-05-27
TASK-528191: fix displaying disabled user in users leaderboard

GouadriaHanen:


ahamdi (changes requested):
We can not filter disabled users using this way, this will cause performance issues depending on the increase of disabled users.

ahamdi (approved):

PR #359, opened 2022-06-10
BUILDERS-194: add ability to edit an announcement activity

GouadriaHanen:


services/src/main/java/org/exoplatform/addons/gamification/listener/challenges/AnnouncementActivityUpdater.java
```diff
@@ -0,0 +1,65 @@
...
+
+  public AnnouncementActivityUpdater(ActivityManager activityManager, AnnouncementService announcementService) {
+    this.activityManager = activityManager;
+    this.announcementService = announcementService;
+  }
+
+  @Override
+  public void updateActivity(ActivityLifeCycleEvent activityLifeCycleEvent) {
+    ExoSocialActivity activity = activityLifeCycleEvent.getSource();
+    if (!activity.getType().equals(ANNOUNCEMENT_ACTIVITY_TYPE)) {
```
boubaker: the `activity.getType()` can be null, thus use this util class from apache commons lang package:
```suggestion
    if (!StringUtils.equals(activity.getType(), ANNOUNCEMENT_ACTIVITY_TYPE)) {
```
or you can at least too, use it this way to ensure to not have an NPE (constant before) :
```suggestion
    if (!ANNOUNCEMENT_ACTIVITY_TYPE.equals(activity.getType())) {
```

services/src/main/java/org/exoplatform/addons/gamification/listener/challenges/AnnouncementActivityUpdater.java
```diff
@@ -0,0 +1,65 @@
...
+    long announcementId = Long.parseLong(activity.getTemplateParams().get("announcementId"));
+    try {
+      Announcement announcement = announcementService.getAnnouncementById(announcementId);
+      if (announcement != null) {
+        announcement.setComment(activity.getTitle());
+        announcementService.updateAnnouncement(announcement);
+        activity.getTemplateParams().replace("announcementComment", activity.getTitle());
+        activityManager.updateActivity(activity, false);
+      }
+    } catch (ObjectNotFoundException e) {
```
boubaker: the `ObjectNotFoundException` isn't thrown by `announcementService.getAnnouncementById`, thus no need to catch it here

services/src/main/java/org/exoplatform/addons/gamification/listener/challenges/AnnouncementActivityUpdater.java
```diff
@@ -0,0 +1,65 @@
...
+  @Override
+  public void updateActivity(ActivityLifeCycleEvent activityLifeCycleEvent) {
+    ExoSocialActivity activity = activityLifeCycleEvent.getSource();
+    if (!activity.getType().equals(ANNOUNCEMENT_ACTIVITY_TYPE)) {
+      return;
+    }
+    long announcementId = Long.parseLong(activity.getTemplateParams().get("announcementId"));
+    try {
+      Announcement announcement = announcementService.getAnnouncementById(announcementId);
+      if (announcement != null) {
```
boubaker:
```suggestion
      if (announcement == null) {
        LOG.warn("Announcement with id {} wasn't found, only the activity message will be updated", announcementId, e);
      } else {
```

services/src/main/java/org/exoplatform/addons/gamification/listener/challenges/AnnouncementActivityUpdater.java
```diff
@@ -0,0 +1,65 @@
...
+        announcement.setComment(activity.getTitle());
+        announcementService.updateAnnouncement(announcement);
+        activity.getTemplateParams().replace("announcementComment", activity.getTitle());
+        activityManager.updateActivity(activity, false);
```
boubaker: We shouldn't duplicate the announcement content inside a template param. Why haven't we used the activity.title to put user message ? (I know that in kudos, it was added like this, but it was because of extensibility limitation due the rigidity of previous development process)

services/src/main/java/org/exoplatform/addons/gamification/listener/challenges/AnnouncementActivityUpdater.java
GouadriaHanen: done!

services/src/main/java/org/exoplatform/addons/gamification/listener/challenges/AnnouncementActivityUpdater.java
GouadriaHanen: refactored!

services/src/main/java/org/exoplatform/addons/gamification/listener/challenges/AnnouncementActivityUpdater.java
GouadriaHanen: done!

services/src/main/java/org/exoplatform/addons/gamification/listener/challenges/AnnouncementActivityUpdater.java
GouadriaHanen: it is not duplicating it is just replacing the old value!

services/src/main/java/org/exoplatform/addons/gamification/listener/challenges/AnnouncementActivityUpdater.java
```diff
@@ -0,0 +1,66 @@
...
+  public void updateActivity(ActivityLifeCycleEvent activityLifeCycleEvent) {
+    ExoSocialActivity activity = activityLifeCycleEvent.getSource();
+    if (!StringUtils.equals(activity.getType(), ANNOUNCEMENT_ACTIVITY_TYPE)) {
+      return;
+    }
+    long announcementId = Long.parseLong(activity.getTemplateParams().get("announcementId"));
+    Announcement announcement = announcementService.getAnnouncementById(announcementId);
+
+    try {
+      if (announcement != null) {
```
boubaker, line 55: No need to check on nullability if you catch the `ObjectNotFoundException` when the object is not found. In addition the try/catch has to be made with `    Announcement announcement = announcementService.getAnnouncementById(announcementId);` inside it, not before the `try/catch` block.

services/src/main/java/org/exoplatform/addons/gamification/listener/challenges/AnnouncementActivityUpdater.java
```diff
@@ -0,0 +1,66 @@
...
+        announcement.setComment(activity.getTitle());
+        announcementService.updateAnnouncement(announcement);
+      }
+    } catch (ObjectNotFoundException e) {
+      LOG.warn("Announcement with id {} wasn't found, only the activity message will be updated", announcementId, e);
+    }
+    activity.getTemplateParams().replace("announcementComment", activity.getTitle());
```
boubaker: where is used the `announcementComment` ?

boubaker: ok changed by commited https://github.com/Meeds-io/gamification/pull/359/commits/0691cb90451952854de71198b71f0c11624c5ae7 as discussed

boubaker (approved):

PR #361, opened 2022-06-14
BUILDERS-265: Reuse the Activity CKeditor in announcement drawer

GouadriaHanen:


challenges-webapp/src/main/webapp/vue-app/challenges/challengesServices.js
```diff
@@ -74,7 +74,7 @@ export function getAllAnnouncementsByChallenge(challengeId, offset, limit) {
 }
 
 export function saveAnnouncement(announcement) {
-  return fetch(`${eXo.env.portal.context}/${eXo.env.portal.rest}/gamification/announcement/api/addAnnouncement`, {
+  return fetch(`${eXo.env.portal.context}/${eXo.env.portal.rest}/gamification/announcement/api/`, {
```
boubaker:
```suggestion
  return fetch(`${eXo.env.portal.context}/${eXo.env.portal.rest}/challenges/announcements`, {
```

challenges-webapp/src/main/webapp/vue-app/challenges/components/AnnouncementDrawer.vue
```diff
@@ -42,13 +42,21 @@
...
       <div class="pl-4 pr-4 pt-9 py-4 my-2">
-        <challenge-description
-          ref="challengeDescription"
-          v-model="announcement.comment"
-          :value="announcement.comment"
-          @invalidDescription="invalidDescription($event)"
-          @validDescription="validDescription($event)"
-          @addDescription="addDescription($event)" />
+        <div
+          id="descriptionId"
```
boubaker: use a more specific identifier

challenges-webapp/src/main/webapp/vue-app/challenges/components/AnnouncementDrawer.vue
```diff
@@ -101,27 +109,32 @@ export default {
...
+      MAX_LENGTH: 1300,
     };
   },
   computed: {
     space() {
       return this.challenge && this.challenge.space;
     },
     disabledSave() {
-      return this.announcement.assignee && this.announcement.assignee.length > 0 && this.isValidDescription.description ;
+      return this.announcement.assignee && this.announcement.assignee.length > 0 && this.validInput && this.announcement.comment && this.announcement.comment.length > 0;
```
boubaker, line 119: we may indicate to the end user what field is mandatory and must be filled, else, we will have a very bad UX

challenges-webapp/src/main/webapp/vue-app/challenges/components/AnnouncementDrawer.vue
```diff
@@ -101,27 +109,32 @@ export default {
...
     },
     enableSuggester() {
       return this.challenge && (this.challenge.userInfo && this.challenge.userInfo.manager || this.challenge.userInfo && this.challenge.userInfo.redactor);
     },
     disableSuggester(){
       return this.challenge && this.challenge.userInfo && !this.challenge.userInfo.manager && !this.challenge.userInfo.redactor && this.challenge.userInfo.member;
     },
+    ckEditorId() {
```
boubaker: why should we specify a different instance of ckeditor, knowing that it's the same Form and the same drawer ?

challenges-webapp/src/main/webapp/vue-app/challenges/components/ChallengeDescription.vue
```diff
@@ -48,11 +44,8 @@ export default {
     };
   },
   computed: {
-    title() {
-      return this.isChallenge ? this.$t('challenges.label.describeYourChallenge') : this.$t('challenges.label.describeYourAchievement');
-    },
     challengeDescription () {
-      return this.isChallenge && this.challenge && this.challenge.id && this.challenge.description || '';
+      return this.challenge && this.challenge.id && this.challenge.description || '';
```
boubaker:
```suggestion
      return this.challenge?.description || '';
```

services/src/main/java/org/exoplatform/addons/gamification/service/AnnouncementService.java
```diff
@@ -31,7 +32,7 @@ List<Announcement> findAllAnnouncementByChallenge(long challengeId, int offset,
    * @throws IllegalAccessException when user is not authorized to create a
    *           announcement for the designated owner defined in object
    */
-  Announcement createAnnouncement(Announcement announcement, String username, boolean system) throws IllegalAccessException,
+  Announcement createAnnouncement(Announcement announcement, Map<String, String> templateParams , String username, boolean system) throws IllegalAccessException,
```
boubaker, line 35:
```suggestion
  Announcement createAnnouncement(Announcement announcement, String username, boolean system) throws IllegalAccessException,
```
You can define another DTO (ActivityAnnouncement) that inherits from `Announcement` that defines templateParams as a new attribute.

challenges-webapp/src/main/webapp/vue-app/challenges/components/AnnouncementDrawer.vue
```diff
@@ -18,7 +18,7 @@
       </div>
       <hr class="ml-4 mr-4 separator">
       <div class="pl-4 pr-4 mt-7 descriptionLabel">
-        {{ $t('challenges.label.assignedAchievement') }}
+        * {{ $t('challenges.label.assignedAchievement') }}
```
boubaker:
```suggestion
        {{ $t('challenges.label.assignedAchievement') }} *
```
The mandatory indication goes after the field name.

challenges-webapp/src/main/webapp/vue-app/challenges/components/AnnouncementDrawer.vue
```diff
@@ -42,13 +42,17 @@
         </v-chip>
       </div>
       <div class="pl-4 pr-4 pt-9 py-4 my-2">
-        <challenge-description
-          ref="challengeDescription"
+        <div class="py-1 px-2 subtitle-1">
+          * {{ $t('challenges.label.describeYourAchievement') }}
```
boubaker:
```suggestion
          {{ $t('challenges.label.describeYourAchievement') }} *
```
The mandatory tip goes after the field name.

challenges-webapp/src/main/webapp/vue-app/challenges/components/AnnouncementDrawer.vue
```diff
@@ -42,13 +42,17 @@
...
+        </div>
+        <exo-activity-rich-editor
           v-model="announcement.comment"
-          :value="announcement.comment"
-          @invalidDescription="invalidDescription($event)"
-          @validDescription="validDescription($event)"
-          @addDescription="addDescription($event)" />
+          ref="ckEditorId"
```
boubaker:
```suggestion
          ref="announcementRichEditor"
```
use a meaningful ref name instead of using a variable name pattern.

challenges-webapp/src/main/webapp/vue-app/challenges/components/ChallengeCard.vue
```diff
@@ -88,7 +88,7 @@
             class="btn btnAdd mx-1"
             :disabled="!enableAnnounce"
             :title="showMessage"
-            @click="createAnnounce">
+            @click="$emit('create-announce', challenge )">
```
boubaker:
```suggestion
            @click="$root.$emit('open-announcement-drawer', challenge )">
```

challenges-webapp/src/main/webapp/vue-app/challenges/components/ChallengeCard.vue
```diff
@@ -173,23 +169,23 @@ export default {
...
     getListWinners() {
       this.challenge.announcements.map(announce => {
         this.listWinners.push({'userName': announce.assignee});
       });
     },
-    announcementAdded(announcement) {
-      this.listWinners.unshift({'userName': announcement.assignee});
-      this.challenge.announcementsCount = this.challenge.announcementsCount + 1;
+    announcementAdded(event) {
+      if (event && event.detail) {
```
boubaker, line 181:
```suggestion
      if (event?.detail) {
```

challenges-webapp/src/main/webapp/vue-app/challenges/components/ChallengeCard.vue
```diff
@@ -173,23 +169,23 @@ export default {
...
+        this.listWinners.unshift({'userName': event.detail.assignee});
+        this.challenge.announcementsCount = this.challenge.announcementsCount + 1;
+      }
```
boubaker:
```suggestion
      const announcement = event?.detail;
      if (announcement) {
        this.listWinners.unshift({
          userName: announcement.assignee
        });
        this.challenge.announcementsCount++;
      }
```

challenges-webapp/src/main/webapp/vue-app/challenges/components/Challenges.vue
```diff
@@ -20,7 +20,10 @@
     </v-toolbar>
     <template v-if="displayChallenges">
       <div class="pl-2 pt-5">
-        <challenges-list :challenges="challenges" @edit-challenge="editChallenge($event)" />
+        <challenges-list
+          :challenges="challenges"
+          @edit-challenge="editChallenge($event)"
+          @create-announce="createAnnounce($event)" />
```
boubaker:
```suggestion
          @create-announce="createAnnounce($event)" />
```
We must delete `createAnnounce` and make the drawer handling its events inside it with `this.$root.$on('open-announcement-drawer'` and `document.addEventListener('open-announcement-drawer'`

challenges-webapp/src/main/webapp/vue-app/challenges/components/ChallengeCard.vue
```diff
@@ -88,7 +88,7 @@
             class="btn btnAdd mx-1"
             :disabled="!enableAnnounce"
             :title="showMessage"
-            @click="createAnnounce">
+            @click="$root.$emit('open-announcement-drawer', challenge )">
```
boubaker:
```suggestion
            @click="$root.$emit('open-announcement-drawer', challenge)">
```

challenges-webapp/src/main/webapp/vue-app/challenges/components/ChallengeCard.vue
```diff
@@ -173,23 +169,27 @@ export default {
   },
   created() {
     this.getListWinners();
+    document.addEventListener('announcement-added', this.announcementAdded);
```
boubaker: whether use `$root.$on` everywhere to allow events to be catched/triggered inside the application only, or use `document.addEventListener` when it's a cross applications components. In this specific case, it's better to use `$root.$on` because, this will not be catched outside the application

challenges-webapp/src/main/webapp/vue-app/challenges/components/ChallengesList.vue
```diff
@@ -10,7 +10,10 @@
           md="6"
           lg="4"
           xl="3">
-          <challenge-card :challenge="challenge" @edit="$emit('edit-challenge',challenge)" />
+          <challenge-card
+            :challenge="challenge"
+            @edit="$emit('edit-challenge',challenge)"
+            @create-announce="$emit('create-announce',challenge)" />
```
boubaker: this is not needed anymore, isn't it, we may need to open the drawer from the card itself using `this.$root.$emit('open-announcement-drawer', challenge)` inside `<challenge-card`

services/src/main/java/org/exoplatform/addons/gamification/listener/challenges/AnnouncementActivityGeneratorListener.java
```diff
@@ -68,12 +87,14 @@ public void onEvent(Event<AnnouncementService, Announcement> event) throws Objec
     }
   }
 
-  private ExoSocialActivity createActivity(Announcement announcement, Challenge challenge) throws ObjectNotFoundException {
+  private ExoSocialActivity createActivity(Announcement announcement, Map<String, String> params, Challenge challenge) throws ObjectNotFoundException {
+    if (params == null) {
+      params = new HashMap<>();
+    }
```
boubaker, line 93: No need to instantiate template params with an empty map if it's null

```suggestion
    if (params != null) {
      activity.setTemplateParams(params);
    }
```

services/src/main/java/org/exoplatform/addons/gamification/service/EntityBuilder.java
```diff
@@ -0,0 +1,52 @@
...
+package org.exoplatform.addons.gamification.service;
+
+import org.exoplatform.addons.gamification.service.dto.configuration.Announcement;
+import org.exoplatform.addons.gamification.service.dto.configuration.AnnouncementActivity;
+
+import java.util.Map;
+
+public class EntityBuilder {
+
+  public EntityBuilder() { //NOSONAR
```
boubaker:
```suggestion
  private EntityBuilder() {
    // Class having static members only, thus default constructor is declared as private 
```

services/src/main/java/org/exoplatform/addons/gamification/service/dto/configuration/AnnouncementActivity.java
```diff
@@ -0,0 +1,129 @@
...
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+package org.exoplatform.addons.gamification.service.dto.configuration;
+
+import java.util.Map;
+
+public class AnnouncementActivity implements Cloneable {
```
boubaker, line 27: Use lombok

```suggestion
@Data
public class AnnouncementActivity implements Cloneable {
```

services/src/main/java/org/exoplatform/addons/gamification/utils/Utils.java
```diff
@@ -101,17 +101,16 @@ public static final boolean canEditChallenge(List<Long> managersId, String space
     return isChallengeOwner && isSpaceManager;
   }
 
-  public static final boolean canAnnounce(String id) {
+  public static final boolean canAnnounce(String spaceId, String username) {
     SpaceService spaceService = CommonsUtils.getService(SpaceService.class);
```
boubaker, line 105: It would be better to not retrieve services statically but add depending services as parameter inside the method signature to make sure that the caller Service has a dependency on it. (This should be used only by components that aren't managed by IOC)
( This change is optional, and just for best practices knowledge share )

services/src/main/java/org/exoplatform/addons/gamification/listener/challenges/AnnouncementActivityGeneratorListener.java
GouadriaHanen: this check is added in case no templeteParams is sent when it is null
so I can add extra params

services/src/main/java/org/exoplatform/addons/gamification/utils/Utils.java
GouadriaHanen: well understood! thank you

services/src/test/java/org/exoplatform/addons/gamification/listener/AnnouncementActivityGeneratorListenerTest.java
```diff
@@ -0,0 +1,116 @@
+package org.exoplatform.addons.gamification.listener;
```
boubaker, line 17: license header

GouadriaHanen: done!

boubaker (approved):

PR #362, opened 2022-06-14
 BUILDERS-194: add ability to edit an announcement activity

GouadriaHanen:


boubaker:
you have a Sonar check fail

services/src/main/java/org/exoplatform/addons/gamification/listener/challenges/AnnouncementActivityUpdater.java
```diff
@@ -0,0 +1,59 @@
+/*
```
boubaker, line 1:
```suggestion
/**
```

boubaker (approved):

PR #368, opened 2022-06-20
Task-55923: Add gamification label to gamification en_properties

sofyenne:
ISSUE : in gamification earn points some inputs of agenda rules are displayed in  English when user change the language to french 
FIX : Add rule gamification label to gamification en_properties


services/src/main/resources/locale/addon/Gamification_en.properties
```diff
@@ -23,6 +23,16 @@ exoplatform.gamification.leaderboard.rank=Your current rank
 exoplatform.gamification.leaderboard.Pie=Pie chart loading
 exoplatform.gamification.leaderboard.Pieopen=Open Pie chart
 #Gamification Informations Portlet#
+exoplatform.gamification.gamificationinformation.rule.title.CreateEvent=Agenda : Create an event
```
boubaker, line 26: no agenda in Meeds-io, so you may need to add those labels in agenda instead of here (with the same path to extend its content)

PR #374, opened 2022-06-23
Task-58487: Challenge : can't add any challenge in case of migration from 6.2.5

hchaabene:
Problem: when migrate 6.2 to 6.3 ,error displayed in log terminal , that indicate duplicated key in GAMIFICATION_RULE , so can't add challenge , auto Increment not working.
Fix: i remove selectror <addAutoIncrement for column id , because it's duplicated declaration for  autoIncrement ID mention in table  GAMIFICATION_RULE.

### DIFF
--- a/services/src/main/resources/db/changelog/gamification.db.changelog-1.0.0.xml
+++ b/services/src/main/resources/db/changelog/gamification.db.changelog-1.0.0.xml
@@ -437,4 +437,15 @@ Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
                           tableName="GAMIFICATION_ACTIONS_HISTORY"/>
     </changeSet>
 
+    <changeSet id="1.0.0-38" author="exo-gamification">
+        <sql
+                dbms="mysql"
+                endDelimiter=";"
+                splitStatements="true"
+                stripComments="true">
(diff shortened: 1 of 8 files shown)


services/src/main/resources/db/changelog/gamification.db.changelog-1.0.0.xml
```diff
@@ -410,8 +410,6 @@ Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
         </createIndex>
     </changeSet>
     <changeSet id="1.0.0-35" author="exo-gamification">
-        <addAutoIncrement columnDataType="BIGINT" columnName="ID" incrementBy="1" startWith="1"
-                          tableName="GAMIFICATION_RULE"/>
```
ahamdi: We can not change liquibase changelogs this way

ahamdi (changes requested):

ahamdi:
@hchaabene could you use SQL with liquibase to fix the initial ID of the sequence of primary key of the table GAMIFICATION_RULE? 
https://dev.mysql.com/doc/refman/5.7/en/information-functions.html#function_last-insert-id

ahamdi (approved):

PR #376, opened 2022-06-23
MEED-295 Adapt Number of displayed challenges switch Screen resolution

MayTekayaa:
Prior to this change, the retrieved challenges count are the same independently from screen resolution. Consequently, the number of rows can vary switch screen resolution. 
This change allows to display only 2 lines of challenges independently from screen resolution.

### DIFF
--- a/challenges-webapp/src/main/webapp/skin/less/challenges.less
+++ b/challenges-webapp/src/main/webapp/skin/less/challenges.less
@@ -267,6 +267,30 @@
 .emptyChallenges {
   min-height: ~"calc(100vh - 100px)" ;
 }
+@media (max-width : @maxMobileWidth) {
+  .challenge-column {
+    width: 100%;
+    
+    }
+}
+@media (min-width : @minTabletWidth) and (max-width: @maxLargeTabletWidth) {
+  .challenge-column {
+    width: 50%;
+    
+    }
+}
+@media (min-width : @minDesktop) and (max-width: 1365px) {
+  .challenge-column {
+    width: 25%;
+    
+    }
+}
+@media (min-width : 1366px) {
+  .challenge-column {
+    width: 17%;
+    
+    }
+}
 .challenge-card {
   .cardOfChallenge {
     width: 95%;
--- a/challenges-webapp/src/main/webapp/skin/less/challenges.less
+++ b/challenges-webapp/src/main/webapp/skin/less/challenges.less
@@ -267,30 +267,6 @@
 .emptyChallenges {
   min-height: ~"calc(100vh - 100px)" ;
 }
-@media (max-width : @maxMobileWidth) {
-  .challenge-column {
-    width: 100%;
-    
-    }
-}
-@media (min-width : @minTabletWidth) and (max-width: @maxLargeTabletWidth) {
-  .challenge-column {
-    width: 50%;
-    
-    }
-}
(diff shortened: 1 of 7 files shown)


challenges-webapp/src/main/webapp/vue-app/challenges/components/DomainChallengesList.vue
```diff
@@ -37,20 +37,23 @@ Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
...
-            xl="3">
-            <challenge-card
-              :domain="domain"
-              :challenge="challenge" />
-          </v-col>
+          <template v-for="(challenge, index) in challenges">
+            <v-col
+              :key="index"
+              class="mb-4 challenge-column"
+              >
```
boubaker:
```suggestion
            <v-col
              v-for="challenge in challenges"
              :key="challenge.id"
              class="mb-4 challenge-column">
```

services/src/main/java/org/exoplatform/addons/gamification/storage/cached/RuleCachedStorage.java
```diff
@@ -102,9 +102,8 @@ public List<RuleEntity> findAllChallengesByUser(int offset, int limit, List<Long
 
   @Override
   public List<RuleEntity> findAllChallengesByUserByDomain(long domainId, int offset, int limit, List<Long> ids) {
-    return (List<RuleEntity>) this.ruleFutureCache.get(new CacheKey(CHALLENGE_USER_DOMAIN_CONTEXT, ids, domainId, offset, limit),
-                                                       CHALLENGE_USER_DOMAIN_CONTEXT + domainId + offset
-                                                           + Utils.getCurrentUser());
+    CacheKey key = new CacheKey(CHALLENGE_USER_DOMAIN_CONTEXT, ids, domainId, offset, limit);
+    return (List<RuleEntity>) this.ruleFutureCache.get(key, key.hashCode());
```
boubaker, line 109: To apply this change in all other methods using `CacheKey` as well please. If we discovered a bug using this notation, it's better to fix it in all locations at the same time and avoid waiting dedicated tasks to fix it one by one.

challenges-webapp/src/main/webapp/skin/less/challenges.less
```diff
@@ -267,6 +267,30 @@
...
+    width: 25%;
+    
+    }
+}
+@media (min-width : 1366px) {
+  .challenge-column {
+    width: 17%;
+    
+    }
+}
```
boubaker: Using `col-` classes, we don't need those CSS changes anymore.

challenges-webapp/src/main/webapp/vue-app/challenges/components/DomainChallengesList.vue
MayTekayaa: the v-for in template is used to display the challenges as indicated in the US, so if there are 8 to be displayed, we will have 4 in a first line and four others in a second one

challenges-webapp/src/main/webapp/skin/less/challenges.less
MayTekayaa: col- does not meet our needs, I need to display the number of challenges indicated (4 challenges in the first row) on the width of the whole interface

boubaker (approved):

PR #390, opened 2022-06-30
MEED-257 Add programs tab to challenges application

MayTekayaa:
This feature adds a navbar at the top of the engagement center, when feature flag is enabled, to allow the user to access to programs. When the feature flag is disabled, the navbar isn't displayed and the Challenges application display remains unchanged.


challenges-webapp/src/main/webapp/vue-app/challenges/components/EngagementCenter.vue
```diff
@@ -0,0 +1,55 @@
+<!--
+This file is part of the Meeds project (https://meeds.io/).
+Copyright (C) 2020 Meeds Association
+contact@meeds.io
```
plamarque: 2022

challenges-webapp/src/main/resources/locale/portlet/Challenges_en.properties
```diff
@@ -14,6 +14,9 @@
 # along with this program; if not, write to the Free Software Foundation,
 # Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
 #
+engagementCenter.label.realizations=Realizations
```
plamarque: should rather be Challenges.

plamarque (changes requested):

challenges-webapp/src/main/webapp/skin/less/challenges.less
```diff
@@ -1,10 +1,13 @@
 @import "variables.less";
 @import "mixins.less";
 
-#ChallengesApplication {
+#EngagementCenterApplication {
```
boubaker, line 4: You may not need this change, please the ID `ChallengesApplication` inside `Challenges.vue`

MayTekayaa: I have to do it because I have to set the margin for the whole interface, I put challenge-applications for the challenge style

challenges-webapp/src/main/webapp/vue-app/challenges/components/EngagementCenter.vue
MayTekayaa: done

challenges-webapp/src/main/resources/locale/portlet/Challenges_en.properties
MayTekayaa: done

challenges-webapp/src/main/webapp/skin/less/challenges.less
boubaker: as discussed, this should be managed to keep the exact structure that was before by applying the ID on the newly introduced parent element (`div` instead of `v-app`) inside `Challenges.vue`. This will avoid to change the HTML structure of the page. For the margin, it would be better to apply it on the toolbar so that it's added only when the toolbar is visible (Feature flag enabled).

challenges-webapp/src/main/webapp/vue-app/engagement-center/components/challenges/Challenges.vue
```diff
@@ -1,5 +1,6 @@
 <template>
-  <v-app
+  <div
+  id="ChallengesApplication"
```
boubaker:
```suggestion
    id="ChallengesApplication"
```

MayTekayaa: fixed

boubaker (approved):

plamarque (approved):
LGTM

PR #393, opened 2022-07-06
MEED-196: fix searching challenges

GouadriaHanen:
> prior to this change: 
> 
> - searching challenges is only with the exact words not with keywords
> - announcement activity id is null 
> - rule filtring false results

> after this change:
> 
> - search challenge is able with at least 3 letters  for title and exact word for the description
> - announcement activity id
> - rule filtering displays well result

### DIFF
--- a/challenges-webapp/src/main/webapp/vue-app/engagement-center/components/challenges/Challenges.vue
+++ b/challenges-webapp/src/main/webapp/vue-app/engagement-center/components/challenges/Challenges.vue
@@ -35,18 +35,18 @@ Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
       </div>
       <v-spacer />
       <div class="challengeFilter text-center d-flex align-center justify-space-around">
-        <v-tooltip bottom>
-          <template #activator="{ on }">
-            <v-text-field
-              id="EngagementCenterApplicationSearchFilter"
-              v-on="on"
-              v-model="search"
-              :placeholder="$t('challenges.filter.search')"
-              prepend-inner-icon="fa-filter"
-              single-line
-              hide-details
-              class="pa-0 mx-3" />
-          </template>
+        <v-text-field
+          id="EngagementCenterApplicationSearchFilter"
+          v-model="search"
+          :placeholder="$t('challenges.filter.search')"
+          prepend-inner-icon="fa-filter"
+          single-line
+          hide-details
+          class="pa-0 mx-3" />
+        <v-tooltip
+          id="EngagementCenterApplicationSearchFilterTooltip"
+          :value="displayMinimumCharactersToolTip"
+          attach>
           <span> {{ $t('challenges.filter.searchTooltip') }} </span>
         </v-tooltip>
       </div>
@@ -60,7 +60,7 @@ Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
       :domains="domainsHavingChallenges"
       :challenges-by-domain-id="challengesByDomainId"
       :loading="loading"
-      class="pl-2 pt-5" />
+      class="pl-2 pt-8" />
 
     <challenge-drawer v-if="canAddChallenge" ref="challengeDrawer" />
     <challenge-details-drawer ref="challengeDetails" />
@@ -90,6 +90,7 @@ export default {
     endTypingKeywordTimeout: 50,
     startTypingKeywordTimeout: 0,
     typing: false,
+    displayMinimumCharactersToolTip: false,
   }),
   computed: {
     classWelcomeMessage() {
@@ -135,10 +136,13 @@ export default {
   },
   watch: {
     search()  {
-      if (!this.search?.length ) {
+      this.displayMinimumCharactersToolTip = false;
+      if (!this.search?.length) {
         this.getChallenges(false);
-      } else if (this.search.length < 3 ) {
-        return ;
+        return;
+      } else if (this.search.length < 3) {
+        this.displayMinimumCharactersToolTip = true;
+        return;
       }
       this.startTypingKeywordTimeout = Date.now() + this.startSearchAfterInMilliseconds;
       if (!this.typing) {
--- a/challenges-webapp/src/main/webapp/vue-app/engagement-center/components/challenges/Challenges.vue
+++ b/challenges-webapp/src/main/webapp/vue-app/engagement-center/components/challenges/Challenges.vue
@@ -34,15 +34,21 @@ Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
         </v-btn>
       </div>
       <v-spacer />
-      <div class="challengeFilter">
-        <v-text-field
-          id="filter"
-          v-model="search"
-          :placeholder="$t('challenges.filter.search')"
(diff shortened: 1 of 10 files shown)


challenges-webapp/src/main/webapp/vue-app/engagement-center/components/challenges/Challenges.vue
```diff
@@ -20,6 +20,7 @@
       <v-spacer />
       <div class="challengeFilter">
         <v-text-field
+          id="filter"
```
boubaker: more specific id like, see https://community.exoplatform.com/portal/dw/activity?id=399725

extension/src/main/webapp/WEB-INF/conf/gamification/search/rule-search-query.json
```diff
@@ -14,7 +14,7 @@
         {
           "query_string": {
             "fields": ["title", "description"],
-            "query": "@term_query@",
+            "query": "*@term_query@*",
```
boubaker:
```suggestion
            "query": "@term_query@",
```
Nope, we shouldn't make the search so permissive to be able to search with a letter (like we do with other indices)

services/src/main/java/org/exoplatform/addons/gamification/search/RuleSearchConnector.java
```diff
@@ -60,7 +60,9 @@ public class RuleSearchConnector {
       + "            }\n"
       + "          }\n"
       + "        }\n";
-
+  
+  private static final String          ILLEGAL_SEARCH_CHARACTERS    = "\\!?^()+-=<>{}[]:\"'*~&|";
```
boubaker, line 64: we may have to espace those characters instead of deleting them

boubaker: or you can reuse the same algorith used here for more coherence : https://github.com/Meeds-io/social/blob/b8a321d0711522990b33d2fa342f9a8557fed9af/component/core/src/main/java/org/exoplatform/social/core/jpa/search/ActivitySearchConnector.java#L317-L321

GouadriaHanen: we escape them :
https://github.com/Meeds-io/gamification/pull/393/files#diff-50e9310b0f0ce4e8a8e1f71e32d23a7b3daf42f7f8a7fdefe852570fdf59e284R289

boubaker: yep, to reuse the same algorithm as mentioned in previous [comment](https://github.com/Meeds-io/gamification/pull/393/files#r914909128)

boubaker: ok I see the point, so you already used this algorithm but you want to delete characters from entered query by user. IMO you should just escape the double quote character and that's fine to keep the special characters searched by user.
See example of usage https://stackoverflow.com/a/66305431

extension/src/main/webapp/WEB-INF/conf/gamification/search/rule-search-query.json
boubaker: In fact if you would like to make the search more permissive, you can change the fuzziness index.

services/src/main/java/org/exoplatform/addons/gamification/search/RuleSearchConnector.java
GouadriaHanen: already escaped not deleted :
https://github.com/Meeds-io/gamification/pull/393/files#diff-50e9310b0f0ce4e8a8e1f71e32d23a7b3daf42f7f8a7fdefe852570fdf59e284R289

extension/src/main/webapp/WEB-INF/conf/gamification/search/rule-search-query.json
```diff
@@ -14,7 +14,7 @@
         {
           "query_string": {
             "fields": ["title", "description"],
-            "query": "@term_query@",
+            "query": "title:*@term_query@* OR description:@term_query@~1",
```
boubaker: 1. what if the `term_query` contains multiple words ? you may should use parenthesis as well
2. For the `fuzziness`, it's already configured in a dedicated property so no need to repeat it in `@term_query@~1`
3. the `~` has to be added in Java part for words that are greater than 4 letters ([See example here](https://github.com/Meeds-io/social/blob/develop/component/core/src/main/java/org/exoplatform/social/core/jpa/search/ActivitySearchConnector.java#L374-L381))

services/src/main/java/org/exoplatform/addons/gamification/search/RuleSearchConnector.java
boubaker: ok, the characters to escape are only https://www.elastic.co/guide/en/elasticsearch/reference/7.17/query-dsl-simple-query-string-query.html#simple-query-string-query-notes

boubaker, line 64: The characters to escape with `\` are listed here (avoid adding others please)
https://www.elastic.co/guide/en/elasticsearch/reference/7.17/query-dsl-simple-query-string-query.html#simple-query-string-query-notes

challenges-webapp/src/main/webapp/vue-app/engagement-center/components/challenges/Challenges.vue
```diff
@@ -112,6 +119,11 @@ export default {
   },
   watch: {
     search()  {
+      if (!this.search?.length ) {
+        this.getChallenges(false);
+      } else if (this.search.length < 3 ) {
+        return ;
+      }
```
boubaker: 1. Code Formatting
2. use a data variable to indicate when to display the tooltip (when user entered less than 3 letters)
3. return when no search keyword, else you will make the search twice

```suggestion
      this.displayMinimumCharactersToolTip = false;
      if (!this.search?.length) {
        this.getChallenges(false);
        return;
      } else if (this.search.length < 3) {
        this.displayMinimumCharactersToolTip = true;
        return;
      }
```

services/src/main/java/org/exoplatform/addons/gamification/search/RuleSearchConnector.java
GouadriaHanen: those are the reserved characters for simple query string 
but we need to escape the reserved characters for query string :
 https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-query-string-query.html#_reserved_characters

extension/src/main/webapp/WEB-INF/conf/gamification/search/rule-search-query.json
GouadriaHanen: done!

challenges-webapp/src/main/webapp/vue-app/engagement-center/components/challenges/Challenges.vue
GouadriaHanen: done!

AzmiTouil (approved):
LGTM

PR #395, opened 2022-07-19
MEED-397: Make challenges App available for externals

GouadriaHanen:
prior to this change, challenges Application is not available for external users
after this change, external users are able to use challengesApp


AzmiTouil (approved):
LGTM

PR #396, opened 2022-07-20
MEED-196: fix searching challenges with accents and speacial characters

GouadriaHanen:
before this change, searching challenges with a keyword that ends with space returns all indexed challenges
after this change, searching challenges with a keyword that ends with space returns well the concerned challenges

### DIFF
--- a/services/src/main/java/org/exoplatform/addons/gamification/search/RuleSearchConnector.java
+++ b/services/src/main/java/org/exoplatform/addons/gamification/search/RuleSearchConnector.java
@@ -129,11 +129,20 @@ public class RuleSearchConnector {
   }
 
   private String buildQueryStatement(RuleFilter filter, long offset, long limit) {
-    String term = buildTerm(filter.getTerm());
-    String termQuery = buildTermQuery(filter.getTerm());
-    if (StringUtils.isBlank(term) && StringUtils.isBlank(termQuery)) {
+    String term = removeSpecialCharacters(filter.getTerm());
+    term = escapeIllegalCharacterInQuery(term);
+    term = term.stripLeading().stripTrailing();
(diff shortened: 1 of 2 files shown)


services/src/main/java/org/exoplatform/addons/gamification/search/RuleSearchConnector.java
```diff
@@ -137,6 +132,8 @@ private String buildQueryStatement(RuleFilter filter, long offset, long limit) {
     if (StringUtils.isBlank(term)) {
       return null;
     }
+    term = term.stripLeading();
+    term = term.stripTrailing();
```
AzmiTouil:
```suggestion
    term = term.trim();
```

services/src/main/java/org/exoplatform/addons/gamification/search/RuleSearchConnector.java
```diff
@@ -52,17 +52,12 @@ public class RuleSearchConnector {
...
 
-  private static final String          DOMAIN_FILTERING_QUERY       = ",\n"
-      + "        {\n"
-      + "          \"term\": {\n"
-      + "            \"domainId\": {\n"
-      + "              \"value\": \"@domainId@\"\n"
-      + "            }\n"
-      + "          }\n"
+  private static final String          DOMAIN_FILTERING_QUERY       = ",\n" + "        {\n" + "          \"term\": {\n"
+      + "            \"domainId\": {\n" + "              \"value\": \"@domainId@\"\n" + "            }\n" + "          }\n"
```
AzmiTouil: Please format the query to be more understandable

AzmiTouil (changes requested):

AzmiTouil (approved):

PR #403, opened 2022-07-26
MEED-416: Display default program when there is no program setted yet

GouadriaHanen:
prior to this change when no program is added, an empty page is displayed
after this change, the default programs are displayed since no program is added


challenges-webapp/src/main/webapp/vue-app/engagement-center/initComponents.js
```diff
@@ -46,6 +49,9 @@ const components = {
   'challenge-program': challengeProgram,
   'engagement-center': EngagementCenter,
   'challenge-no-results': NoResultFound,
+  'programs': Programs,
+  'programs-list': ProgramsList,
+  'program-card': ProgramCard,
```
boubaker: Bad Vue component naming, it should be : 
```suggestion
  'engagement-center-programs': Programs,
  'engagement-center-programs-list': ProgramsList,
  'engagement-center-program-card': ProgramCard,
```

See [DEVSTD-34](https://github.com/Meeds-io/meeds/blob/develop/CODING_GUIDELINES.md#devstd-34) for more details

challenges-webapp/src/main/webapp/vue-app/engagement-center/components/programs/Programs.vue
```diff
@@ -0,0 +1,58 @@
...
+
+<script>
+export default {
+  data() {
+    return {
+      programs: null
+    };
+  },
+  created() {
+    this.getAllPrograms();
```
boubaker:
```suggestion
    this.retrievePrograms();
```

Using `get` in method name means that the method will return the desired object without implicit alteraion of current component attributes (this is confusing with Getters & Setters methods definition). Thus you need to name this method differently, `init`, `retrieveXYZ` ... but not using `get`

challenges-webapp/src/main/webapp/vue-app/engagement-center/components/programs/ProgramCard.vue
```diff
@@ -0,0 +1,67 @@
...
+      type: Object,
+      default: null
+    }
+  },
+  computed: {
+    programCover(){
+      return this.program?.cover || '/challenges/images/program_default_cover_back.png';
+    },
+    programBudget(){
+      return this.$t('programs.budget').replace('{0}',this.program?.budget || 0);
```
boubaker:
```suggestion
      return this.$t('programs.budget', {0:  this.program?.budget || 0});
```

The `{0}` parameter in label is already recognized and interpreted by VueI18N, please [read this](https://kazupon.github.io/vue-i18n/guide/formatting.html#named-formatting)

boubaker (approved):

PR #408, opened 2022-07-28
MEED-196: fix indexing and searching challenges with accent

GouadriaHanen:
prior to this change, challenges in which its description contains accents and special characters are indexed with ASCII code
and they can not be  searched
after this change, challenges are index without being coded and they can be searched

### DIFF
--- a/services/src/main/java/org/exoplatform/addons/gamification/connector/RuleIndexingServiceConnector.java
+++ b/services/src/main/java/org/exoplatform/addons/gamification/connector/RuleIndexingServiceConnector.java
@@ -23,6 +23,7 @@ import java.util.Map;
 import java.util.stream.Collectors;
 
 import org.apache.commons.collections.CollectionUtils;
+import org.apache.commons.lang.StringEscapeUtils;
 import org.apache.commons.lang.StringUtils;
 import org.exoplatform.addons.gamification.entities.domain.configuration.RuleEntity;
 import org.exoplatform.addons.gamification.storage.dao.RuleDAO;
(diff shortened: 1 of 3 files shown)


services/src/main/java/org/exoplatform/addons/gamification/connector/RuleIndexingServiceConnector.java
```diff
@@ -119,4 +120,9 @@ private String getUserIdentityId(String username) {
   private String toMilliSecondsString(Date date) {
     return date != null ? String.valueOf(date.getTime()) : "0";
   }
+
+  private String getDescription(String s) {
+    s = StringEscapeUtils.unescapeHtml(s);
+    return s.replace("<", " <").replace(">", "> ");
```
boubaker: why this change ?

GouadriaHanen: we need to escape the Unicode from the description and put a space between the HTML tags and the string if they are concatenated they will not be searched

boubaker: We may have to add a new analyzer such as https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-htmlstrip-charfilter.html#analysis-htmlstrip-charfilter-analyzer-ex that will include `html_strip` as `char_filter` in the default list of analyzers https://github.com/Meeds-io/commons/blob/develop/commons-search/src/main/resources/es-default-index-settings.json#L23-L30, WDYT ?

GouadriaHanen: to be checked!

boubaker (approved):

PR #414, opened 2022-08-02
 MEED-112: implement add program feature

GouadriaHanen:
prior to this change, only from the programs administration page programs could be added.
after this change, adding programs is available on the engagement center programs tab


portlets/webpack.dev.js
```diff
@@ -21,7 +21,7 @@ const common = require('./webpack.common.js');
 
 module.exports = merge(common, {
   output: {
-      path: `/exo-server/webapps/gamification-portlets/`,
+      path: `/home/exo/work/exoRepos/Reward2FT/meedsBuilders/meeds-community-1.4.x-meedsv2-SNAPSHOT/webapps/gamification-portlets/`,
```
boubaker:
```suggestion
      path: `/exo-server/webapps/gamification-portlets/`,
```

boubaker (approved):

PR #415, opened 2022-08-02
MEED-348: Hidden spaces displayed in popular spaces

AzmiTouil:
Prior this change, an admin user can see hidden spaces in these portlets with the label "Hidden spaces"
Update: Do not display these hidden spaces to non members regardless of whether the user is an admin or not.


SaraBoutej (approved):
LGTM

PR #417, opened 2022-08-09
MEED-301 fixing sorting between dates Realizations table

ArfaouiMohamedAmine:
This `fix` is going to enable `sorting` between `dates` .

### DIFF
--- a/portlets/src/main/webapp/vue-app/realizations/components/Realizations.vue
+++ b/portlets/src/main/webapp/vue-app/realizations/components/Realizations.vue
@@ -40,6 +40,7 @@ Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
       :sort-desc.sync="sortDescending"
       disable-pagination
       hide-default-footer
+      must-sort
       class="mx-6 mt-6 realizationsTable">
       <template slot="item" slot-scope="props">
         <realization-item
@@ -77,7 +78,7 @@ export default {
     limit: 10,
     pageSize: 10,
     loading: true,
-    sortBy: '',
+    sortBy: 'date',
     sortDescending: true,
     limitReached: false,
     toDate: new Date().toISOString(),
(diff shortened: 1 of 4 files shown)


services/src/main/java/org/exoplatform/addons/gamification/entities/domain/effective/GamificationActionsHistory.java
```diff
@@ -166,8 +166,8 @@
     query = "SELECT DISTINCT g FROM GamificationActionsHistory g "
         + " WHERE g.earnerType = :type"
         + " AND g.date BETWEEN :fromDate AND :toDate"
-        + " AND (g.ruleId IS NOT NULL AND g.ruleId IN (:ruleIds)) \n"
-        + "      OR (g.actionTitle IS NOT NULL AND g.actionTitle IN (:ruleEventNames)) \n"
+        + " AND (g.ruleId IS NOT NULL AND g.ruleId IN (:ruleIds) \n"
+        + "      OR g.actionTitle IS NOT NULL AND g.actionTitle IN (:ruleEventNames)) \n"
```
boubaker:
```suggestion
        + " AND ((g.ruleId IS NOT NULL AND g.ruleId IN (:ruleIds)) \n"
        + "      OR (g.actionTitle IS NOT NULL AND g.actionTitle IN (:ruleEventNames))) \n"
```

portlets/src/main/webapp/vue-app/realizations/components/Realizations.vue
```diff
@@ -40,7 +40,6 @@ Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
       :sort-desc.sync="sortDescending"
       disable-pagination
       hide-default-footer
-      must-sort
```
boubaker: to keep it as it is to make sure to have the sort option always displayed to end user (default sort = 'date')

boubaker (approved):

PR #424, opened 2022-08-12
MEED-448 Fix test fail on gamification

ArfaouiMohamedAmine:
the test fail is due to the fact that the creation date is the same for all items (CI Job too efficient in such a way that everything is created in the same millisecond)

### DIFF
--- a/services/src/main/java/org/exoplatform/addons/gamification/entities/domain/effective/GamificationActionsHistory.java
+++ b/services/src/main/java/org/exoplatform/addons/gamification/entities/domain/effective/GamificationActionsHistory.java
@@ -145,21 +145,21 @@ import org.exoplatform.commons.api.persistence.ExoEntity;
 )
 @NamedQuery(
     name = "GamificationActionsHistory.findAllAnnouncementByChallenge",
-    query = "SELECT DISTINCT a FROM GamificationActionsHistory a where a.ruleId = :challengeId order by a.createdDate desc"
+    query = "SELECT DISTINCT a FROM GamificationActionsHistory a where a.ruleId = :challengeId order by a.id desc"
 )
 @NamedQuery(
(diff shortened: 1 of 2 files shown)


services/src/main/java/org/exoplatform/addons/gamification/entities/domain/effective/GamificationActionsHistory.java
```diff
@@ -168,8 +168,34 @@
...
+@NamedQuery(
+    name = "GamificationActionsHistory.findRealizationsByDateAscendingUserAccessPrivilege",
+    query = "SELECT DISTINCT g FROM GamificationActionsHistory g"
+        + " WHERE g.earnerType = :type"
+        + " AND g.earnerId = :earnerId"
+        + " AND g.date BETWEEN :fromDate AND :toDate"
+        + " ORDER BY g.id ASC"
+)
+@NamedQuery(
+    name = "GamificationActionsHistory.findRealizationsByDateAndRulesUserAccessPrivilege",
```
boubaker: Useless query ?

ArfaouiMohamedAmine: actually this query has been named to ByEarnerAndDateAndRules
fixed in the next commit

ArfaouiMohamedAmine: wrong commit cherry-picked actually . 
it should be fixed by now

boubaker (approved):

ArfaouiMohamedAmine:
this change is going to fix he test fail which is due to the fact that the creation date is the same for all items (CI Job too efficient in such a way that everything is created in the same millisecond)

PR #433, opened 2022-08-29
MEED-195 disable portlet view caching

ArfaouiMohamedAmine:
this change is going to desable caching for the realizations view since it's going to be dynamically generated depending on the user type; admin or simple user .


challenges-webapp/src/main/webapp/WEB-INF/portlet.xml
```diff
@@ -13,7 +13,6 @@
     <value>/WEB-INF/jsp/challenges.jsp</value>
     </init-param>
     <expiration-cache>-1</expiration-cache>
```
boubaker:
```suggestion
```

boubaker (approved):

PR #449, opened 2022-09-07
feat: Engagement Center Enabling Achievements Table Sorting by Status column for Regular User / Admin view - MEED-496 -io/meeds#13

ArfaouiMohamedAmine:
This change is going to enable Regular User or Admin to sort the realizations table using the status column in an ascending or descending order .


AzmiTouil:
As we discussed, we must have a dynamic query for all sort type.

PR #469, opened 2022-09-15
fix: fix displaying program field with html tags in achievements - MEED-603 - meeds-io/meeds#13

GouadriaHanen:
prior to this change, when listing or exporting achievements, the program field contains HTML tags,
after this change, the program field is well displayed

### DIFF
--- a/portlets/src/main/webapp/vue-app/realizations/components/RealizationItem.vue
+++ b/portlets/src/main/webapp/vue-app/realizations/components/RealizationItem.vue
@@ -30,7 +30,7 @@ Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
       </span> 
     </td>
     <td class="text-truncate align-center">
-      {{ domainDescription }}
+      <span v-sanitized-html="domainDescription"></span>
     </td>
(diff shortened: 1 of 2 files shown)


AzmiTouil (dismissed):

AzmiTouil (approved):

portlets/src/main/webapp/vue-app/realizations/components/RealizationItem.vue
```diff
@@ -30,7 +30,7 @@ Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
       </span> 
     </td>
     <td class="text-truncate align-center">
-      {{ domainDescription }}
+      <span v-sanitized-html="domainDescription"></span>
     </td>
```
boubaker, line 34: @GouadriaHanen you can instead of addin a new DOM element ( when cumulating all of them, it reduces a lot the page loading performances ), you can add `v-sanitized-html` for `td`.
```html
    <td class="text-truncate align-center" v-sanitized-html="domainDescription">
    </td>

```
@AzmiTouil FYI

PR #470, opened 2022-09-16
feat: Engagement Center  Display 25 item per page in realizations table - MEED-478 - Meeds-io/meeds#13

ArfaouiMohamedAmine:
Prior to this change the realizations table implements` 10 rows` with each call, this change is going to `raise the number of achievements showed per page to 25` .

### DIFF
--- a/services/src/main/java/org/exoplatform/addons/gamification/rest/RealizationsRest.java
+++ b/services/src/main/java/org/exoplatform/addons/gamification/rest/RealizationsRest.java
@@ -87,7 +87,7 @@ public class RealizationsRest implements ResourceContainer {
                                      @QueryParam("offset")
                                      int offset,
                                      @Parameter(description = "Limit of result")
-                                     @DefaultValue("10")
+                                     @DefaultValue("25")
                                      @QueryParam("limit")
                                      int limit,
                                      @Parameter(description = "Response Type")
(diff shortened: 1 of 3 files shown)


AzmiTouil (approved):

services/src/main/java/org/exoplatform/addons/gamification/rest/RealizationsRest.java
```diff
@@ -87,7 +87,7 @@ public Response getAllRealizations(@Context HttpServletRequest httpRequest,
                                      @QueryParam("offset")
                                      int offset,
                                      @Parameter(description = "Limit of result")
-                                     @DefaultValue("10")
+                                     @DefaultValue("25")
```
boubaker, line 90: There is no real default value if you already use this check here :
 ```java
   if (!isXlsx && limit <= 0) {
      return Response.status(Response.Status.BAD_REQUEST).entity("Limit must be positive").build();
    }
```

@AzmiTouil @ArfaouiMohamedAmine this line has to be removed

ArfaouiMohamedAmine: @AzmiTouil @boubaker  done, changes has been applied here https://github.com/Meeds-io/gamification/pull/471/files

PR #472, opened 2022-09-19
fix: Engagement Center wrong selected date borns in realizations table - MEED-605 - Meeds-io/meeds#13

ArfaouiMohamedAmine:
In order to `fit` the wanted achievements into the `selected date terminals`, a `period selector` is used .
Prior to this change some selected terminals tends to get an `out of boundaries dates`, 
This fix is going to include the `timezone differences` factor in the way that the period selector works, by `subtracting` the time difference to put it back at the [Coordinated Universal Time](https://en.wikipedia.org/wiki/Coordinated_Universal_Time) (`UTC`)

### DIFF
--- a/portlets/src/main/webapp/vue-app/realizations/components/SelectPeriod.vue
+++ b/portlets/src/main/webapp/vue-app/realizations/components/SelectPeriod.vue
@@ -149,10 +149,10 @@ export default {
       ];
     },
     fromDate() {
-      return this.value && new Date(this.value.min);
+      return this.value && new Date(this.value.min - (new Date().getTimezoneOffset() * 60000));
     },
     toDate() {
-      return this.value && new Date(this.value.max);
+      return this.value && new Date(this.value.max + (new Date().getTimezoneOffset() * 60000));
     },
     maxDate() {
       return new Date().toLocaleDateString('sv-SV');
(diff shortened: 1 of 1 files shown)


portlets/src/main/webapp/vue-app/realizations/components/SelectPeriod.vue
```diff
@@ -149,10 +149,10 @@ export default {
       ];
     },
     fromDate() {
-      return this.value && new Date(this.value.min);
+      return this.value && new Date(this.value.min - (new Date().getTimezoneOffset() * 60000));
```
boubaker, line 152: Not really convinced by this kind of usage of Time zone. It's better to communicate with Server using RFC-3339 to allow the server to understand the date and used Time zone.

ArfaouiMohamedAmine: @boubaker if am not mistaken, I think that this bug has been resolved by your commit https://github.com/Meeds-io/gamification/commit/0aaf994a84eeb50cfe57b11aaac7761704c61d25 , it's not reproduced in the ACC any more .

boubaker: Ah ok, so no need to this change anymore ? If yes, you can close this IMO.

ArfaouiMohamedAmine: okay thank you .

PR #482, opened 2022-09-22
fix: Fix problem when adding chanllenges or domain after migration from 6.2 - MEED-631

AzmiTouil:
This PR will add a custom changelog to reset auto-increment domain an rule tables


boubaker (approved):

boubaker:
@AzmiTouil can you please backport it into develop branch, this is about a fix not a feature.

PR #511, opened 2022-10-10
fix : In reward administration, the request to compute reward is too long - EXO-60157 - meeds-io/meeds#330

rdenarie:
Before this fix, with a lot of line in gamification history table, the request time for the "compute" request exceed the nginx timeout. This is due to the fact that scores are computed users by users. So there are 1 database request by reward plugin and by user. On tribe, 2*148 request. Each databse requests execute in near 800ms, so the request execution time is too long. This fix add change the compute to have one request per reward plugin, compute score for all users.


azayati (approved):
LGTM

PR #569, opened 2022-11-08
fix: Contribution Center The challenge textual filter is not working - MEED-940

AzmiTouil:
Prior this change, The search in challenge is not working


SaraBoutej (approved):
LGTM

PR #629, opened 2022-11-25
feat: Overview Page Display the programs portlet - MEED-674 - Meeds-io/MIPs#10

MayTekayaa:
This change will add a hover effect on program item in the overview page.


portlets/src/main/webapp/vue-app/commons/components/Widget.vue
```diff
@@ -40,7 +40,11 @@ export default {
     seeAllUrl: {
       type: String,
       default: () => '',
-    }
+    },
+    noXPadding: {
```
AzmiTouil: Define extraClass prop instead.

AzmiTouil (changes requested):

portlets/src/main/webapp/vue-app/commons/components/Widget.vue
MayTekayaa: done!

AzmiTouil (approved):

PR #662, opened 2022-12-02
fix: Overview Page Improve the performance of contributions portlet - MEED-1079 - Meeds-io/MIPs#10 

MayTekayaa:
This change will improve the performance of the contributions portlet in the overview page.

### DIFF
--- a/portlets/src/main/webapp/vue-app/profileStats/components/UserPointsWidget.vue
+++ b/portlets/src/main/webapp/vue-app/profileStats/components/UserPointsWidget.vue
@@ -185,15 +185,16 @@ export default {
   },
   methods: {
     getGamificationPoints() {
-      getGamificationPoints(this.period).then(
-        (data) => {
-          this.userPoints = data.points;
-          if (data.points !== 0) {
-            this.$emit('seeAll', true);
-          }
-          this.option.title[0].subtext = data.points;
-        }
-      );
+      this.$nextTick().then(()=>{
+        getGamificationPoints(this.period).then(
+          (data) => {
+            this.userPoints = data.points;
+            if (data.points !== 0) {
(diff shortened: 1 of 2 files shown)


AzmiTouil (approved):

portlets/src/main/webapp/vue-app/profileStats/components/UserPointsWidget.vue
```diff
@@ -222,10 +223,11 @@ export default {
               this.option.series[0].data.push(serie);
             }
           }
-          this.initChart(this.option);
           this.$emit('loadingData', false);
```
boubaker, line 226: This should go to the finally block @MayTekayaa @ArfaouiMohamedAmine @AzmiTouil 

In all widgets, please consider this strategy (not only here, but in all devs):

```javascript
this.loading = true; // this.$emit('loadingData', true);
PromiseExpression
  .then(() => {
    // Do stuff for successful response .....
    // Like making updates, closing the drawer if the user saved or deleted anything ....
  })
  .catch(e => {
    // Do stuff for error responsding, like for example displaying a v-alert for the user to warn him if he made a PUT/POST/PATCH/DELETE action and that it has failed
    // In this block, the state of forms & drawers & .... should never change
  })
  .finally(() => {
     // This block will be executed all time, thus we don't make changes, just inform user that communicating with server has finished by changing the state of "loading" flag 
     this.loading = false; // this.$emit('loadingData', false);
  })
```

we can do the same thing when using an async method:

```javascript
this.loading = true; // this.$emit('loadingData', true);
try {
   const response = await PromiseExpression;
   // do for success Stuff
} catch (e) {
   // do for error Stuff
} finally {
     this.loading = false; // this.$emit('loadingData', false);
}
```

PR #710, opened 2022-12-16
Merge Meeds-io/MIPs#10

### LINKED ISSUE #10: Overview Page
## Rationale

As the meeds software aims to address the employee recognition market, we need at MeedsLab to improve the onboarding into the platform. Among lots of ideas, we have noticed that most importantly, the first page displayed to users is the stream which is not relavant to our goal, detailed above.
So the gap is we have to work a dedicated page for employee recognition user persona. What we propose is a page that group smartly all the assets the meeds software aims to promote with the story that any user can understand

Mockup there and details in functional requirements:

image

Mobile view:
Same portlets will be displayed when using mobile view. For now, no behaviour (except banner, check below), only make sure portlet are viewable in the same column.

Image

###  Access to the page
This page will be enabled as the first page (aka homepage) only for meedslab customers and meeds software only users. By default, this page will be named "Overview". That way, when meeds software Users access the platform, they should have this overview page display as the homepage and it is positionned as the first page in the left menu.

This page will be displayed for Meeds Packages configuration only as the engagement dashboard. The stakeholders will be able to redefine it in their packages when needed, but not configured by default for them to not affect their default displayed pages.

#### When nothing / Empty page with placeholders
When the user has done nothing yet, and when no program, neither challenges are listed, then we should propose placeholders like this:
empty overview page

Each app proposes to the user to check something with a linked text (green meeds color and bold) with this:
- contributions "this" related to programs tab in contributions center
- reputation peers related to people page
- rewards wallet related to wallet page
- challengers "actions" related to challenge tab in contributions center
- available programs "programs" related to programs tab in contributions center
- don't miss it "actions" related to challenges in contributions center

Depending what the user has done, placeholder remains so we can progressively do actions.
For the contributions part, we should display it only once user has done something different than only logging into the platform.

When loading the page, we should have a loader line for all the page and once it is loaded, we can see the whole page. We shouldn't have something appearing app by app but globally so the experience is nice.

And finally, before detailing each app to display:
- when wallet and perkstore are not enabled,
- then we should display the user stats portlet instead ("welcome back john!")

#### Banner
Banner is displayed above all items. For now, this banner won't be managed by platform host. In the future, we will enhance this banner so it is engaging but not for now.
This banner should be responsive with mobile so it is adjusted as planned in the design (check above): indeed, no description will be displayed when using mobile.
image

#### Reuse of some portlets already used elsewhere: points overview, leaderboard chart

These portlets will be reviewed to be adjusted to the overview page use case AND any improvements added will be as well applied to their current use.

**- Activity:** 
Reuse of the points pie chart (currently displayed when clicking to the points total from the user stat portlet): 
* Back button if displayed from a verso behaviour
* Title: My Contributions
* See all: Aligned to the title to display achievements tab in contributions center
* Subtitle : "Points overview"
* Legend of the pie: aligned so we can see it without overlapping behaviour to the pie when the name is longer than it
* Pie: Current behaviours
+ Change of the colors to be clearer (check design)

image

**- Leaderboard chart:** 
Reuse of the leaderboard chart (currently displayed in the stream when clicking to the rank from the user stat portlet): 
* Back button if displayed from a verso behaviour
* Title: Top challengers
* See all: Aligned to the title to access the people page with leaderboard per program
* No subtitle for now
* Dropdown selector to choose to display a leaderboard per program (by default, set to overview)
* Podium with the three first users depending what we display (by default, the general overview leaderboard)
* If the current user is on the podium, the next three users (from 4 to 6)
* If the current user is not on the podium, then the current user name is bold and we can see it between the before and the next user in the rank (example: I am 7th, then i can see my self between the 6th and the 8th) so I can see how I am positioned
image

#### New portlets using display we already have elsewhere: kudos overview, badge overview, earnings, wallet balance

**- My Reputation:** 
This portlet group two portlets we already have: kudos and badges overview:
* Title: My Reputation
* Kudos portlet as it is displayed currently
* Badges overview with improvements to do: Title style same as the kudos one and title labels changed
Other behaviours related to kudos and badges remain active (like clicking to kudos or to badges to open the related drawer)
image

**- My Rewards:** 
This portlet group:
- two portlets we already have: Wallet overview and Wallet balance 
- and one portlet we need to create

* Title: My Rewards
* See all to access the wallet page
* Wallet overview and wallet balance group in one line
* Perk store item to add:
- Title should depend of the results listed below
- See all to access the perkstore page
- This app should list products the following way: 
** from the more recent created products 
** list the 5 products in a carrousel like documents for exo in stream
These perks card should be hilighted when hovering these
** if no perkstore, then we should propose a placeholder with a cart icon and "Soon, use your token to get product and services
image

#### New portlets 

**- Available programs:** 
This portlet proposes the programs that are available to the user depending what he can see. This one might be used elsewhere for example in a space home page for example. That way, it should be easily configurable for the admin.

* Title: Available programs
* See all to access the contributions center page focused to programs page
* Three programs listed with: icon related to the program cover + program name + reminder of how many point the user can get if he does all the actions in the program
* By default, programs listed are from the most rewarded (with the most points) to the less third less rewarded
* programs are listed to the user depending the audience of the program
* When being an admin, a setting should be proposed from the interface so he can change the name of the portlet and so he can change the program to have: with ability to change to from the most recent to the less one or to a custom list (that way, he can choose which program to list from the programs)

> // FYI this portlet could be used in a future feature so we can display it in a space // * When being in a space, we should display programs related to the audience (space chosen when creating/editing programs). The space host, can access the setting so he can change the title and he can change the program to list

image

**- Don't miss it:** 
This portlet proposes the programs that are available to the user depending what he can see. This one might be used elsewhere for example in a space home page for example. That way, it should be easily configurable for the admin.

* Title: Don't miss it
* See all to access the contributions center page focused to challenges page
* Three challenges listed with: icon (check below for icons) + challenge name + reminder of participants + reminder of how many point the user can get if he does the challenge
* In the case the challenge has not been done yet by anybody, a placeholder should be displayed instead of 0 participants - Be the first to announce
* By default, challenges are listed by three max from the more recent to the less one
* When being an admin, a setting should be proposed from the interface so he can change the name of the portlet and so he can change the challenges to list: with ability to change to from the most recent to the less one or to a custom list (that way, he can choose which program to list from the programs). 
* When the challenge is new, we should have a dedicated icon for this
> // FYI, would be done in a future version //* When being in a space, we should display challenges related to the space (as the challenge audience). All the challenges should be listed there (5 max). The space host, can access the setting so he can change the title and he can change the challenges to list by chosing from the audience and he can change the number of challenges to display

image

### 2.1. Performance requirements
Knowing that this page will aggregate multiple portlets, it's important to consider performances of the display of the page by:
Reducing the number of requests displayed in first page load (avoid loading hidden drawers data only when the user clicks to open the drawer itself)
Avoid increasing common CSS classes and use specific CSS file for each portlet
Avoid increasing JS common modules (commonVueComponents) only when components are re-usable
Pre-load HTTP Requests (CSS & JS files & images & REST API calls)
Use HTTP Header cache when relevant to cache static elements and REST API calls
To give a good UX for end-users, the page has to be loaded within 2.5 ~ 3 seconds at max for a server with a lot of data.

### 2.2. Security requirements
The Overview page should be displayed for external and internal users and modifiable by rewarding administrators only.

### 2.3. Extensibility requirements
Each of 7 blocks will be a portlet apart that will reuse components from other existing portlets. Each portlet will be defined and added into a dedicated category in Application Registry.

### 2.4. Configurability requirements
The page will be displayed by default for Meeds package, but not enabled by default in other packages (eXo). Thus, it should be defined inside the Meeds extension only.

## 3.1. Security
The page has to be allowed for `*:/platform/users` (internal) & `*:/platform/externals` (external) users. The edit permission will be allowed to `*:/platform/rewarding` (Rewarding Administrators) only.

### 3.2.1. Page structure
The Engagement dashboard page will be defined for Meeds Package as the new Home Page. Thus it will be defined in `meeds.war` inside `pages.xml` & `navigation.xml`. Each of **7 blocks** of the page will be a portlet apart. Each portlet will have to inherit the same permissions as page definition for access & edit modes.
The source code of each portlet will be added as follows :

- Banner portlet : `social-portlet.war`
- My Contributions : `gamification-portlets.war`
- My Reputation : `gamification-portlets.war` (with kudos block defined in `kudos.war` and injected in UI through `extension-registry`)
- My Rewards : `gamification-portlets.war` (with kudos block defined in `perk-store.war` and injected in UI through `extension-registry`)
- Top Challengers : `gamification-portlets.war`
- Available Programs : `gamification-portlets.war`
- Don't miss it :  `gamification-portlets.war`

### 3.2.2. Portlets Performances
  To improve the performances of the page loading, the REST calls has to be pre-loaded and thus added in the list URLs to preload by configuration that you can add in :

- `portlet.xml` example : https://github.com/Meeds-io/kudos/commit/2e36d9d48a6698586f70c2b2f930a5484a2dcdd9#diff-30b07608295552d3bc9c1f00e7fb0cbf19400aec4d9543f4de87211be1fc6181R15-R22
- (if not a static URL) use LINK HTTP Header to preload REST calls when possible, like made in ActivityStream : https://github.com/Meeds-io/social/blob/f44bf811eac8764e31ce7573be65a568fac9acf8/component/service/src/main/java/org/exoplatform/social/rest/impl/activity/ActivityRestResourcesV1.java#L950-L958
- Add a `<link preload>` inside JSP file of portlet when the URL is dynamic and could be determined from server end

Each portlet will have a set of Vue components to be displayed. Some of those components comes from other addons. To be able to inject Vue components from other addons, we will use as usual extension-registry.js mechanism. **The drawback of current strategy** is that we will need to register all extensions inside:

- dynamic container inside shared layout => JS files will be loaded in all pages and thus consume more memory, CPU and Network Bandwidth
- dynamic container inside inside the dashboard page => the JS files will not be loaded automatically when adding the portlet outside dashboard page (through layout-management of eXo by example)

To avoid this, we will need to use the notion of JS Modules Group. In fact, currently, this mechanism is used to load **baseGrp** only that unifies all basic JS modules useful for WebUI based portlets. An [improvement can be made](https://github.com/Meeds-io/gatein-portal/commit/fee6d4a1317a8051d0f826d0a2a3874e9f1ef9a0) to make the group definition is allowed cross Web Archives contexts. This way, we will be able to define a group of Modules for each portlet to allow injecting extension into it. For example, the module "[User & Spaces Popover](https://github.com/Meeds-io/social/blob/develop/webapp/portlet/src/main/webapp/WEB-INF/gatein-resources.xml#L1351)" module can have multiple buttons into it and all of those extensions are injected using the Shared layout dynamic container. To change this by the new way, we will have to :

1. Add just `<load-group>popovers</load-group>` inside the definition of [JS Module of Portlet Popover](https://github.com/Meeds-io/social/blob/bc47ca0a8c48d19f337babd4fb37b123455e2f9f/webapp/portlet/src/main/webapp/WEB-INF/gatein-resources.xml#L1350-L1377)
2. Add `<module></module>` for extensions with same `<load-group>popovers</load-group>` for extensions

That's it ! (No more configuration, neither `pages.xml` dynamic container nor a Portlet definition + Kernel configuration to inject a script for Popovers extension).

### 3.2.3. Empty portlets & Loading effect
All portlets will have to load its data through REST calls. Thus, a loading bar has to be added for each block (See [Vuetify v-card loading](https://vuetifyjs.com/en/components/cards/#loading)) until the data is loaded and displayed. When the data is displayed, as usual, the call this.$root.$applicationLoaded() has to be made to signal that overall page has been loaded. The "empty data" block must be displayed while loading.

### 3.2.4. Banner Portlet
The banner portlet will have for now a static image to display, nothing more, so no need to a Vue application, neither a JSP, just a static html file with  inside it. Just couple of things to consider for this image:

- Make sure that the image rendering is cached with an efficient cache time, like 1 year
- Make sure to have a reduced and optimized image size for web rendering, like `WebP` (See https://www.designpowers.com/blog/image-file-formats for more information). This has to be discussed with PO & Designer.

The portlet definition will have to define some `portlet-preferences` (with placeholder as default values) to allow choosing the :

- URL of the image to display
- the description (label Key) to display
- the caption (label Key) to display

By doing this, we will be able to reuse this portlet in other places with different settings only.
### 3.2.5. My Contributions
This new portlet will reuse the same components defined in leaderboard portlet used in stream page. Some UI adaptation will be made to impact this re-usable component, such as Pie Chart color. The loading performances has to be improvement to make this Pie chart displayed on page load. In addition, the Pie chart sometimes isn't well displayed, when the data isn't yet retrieved from server end, thus a stabilization has to be made on it to make sure not processing on chart drawing before retrieving the data.
### 3.2.6. My Reputation
This portlet will have two blocks:

- Kudos block coming from re-usable components already displayed in User Profile Page
- Badges block that will be reused from User Profile Page as well

Both blocks will be injected through extension registry to a common extension name and type. The portlet will have to be defined in gamification codebase (same as badges) and the kudos block has to be defined inside extension in kudos codebase.
### 3.2.7. My Rewards
This portlet will have three blocks:

- Earnings block that is already displayed in User Profile page and that should be reused here through extension-registry
- Wallet Balance that is already displayed in Stream page and that should be reused here through extension-registry
- Perk Store list that should be developed in perk-store code base and that should be reused here through extension-registry

The portlet codebase will be defined in gamification and will have injected extensions to display Perk Store section using a JS Module group loading (See paragraphs above).
For the Perk store section, when the currently authenticated user is a rewarding administrator, an icon will be displayed to let user choose the settings of this portlet to select perk store products to display and/or automatic sort field to use to displays products. This will be a portlet preference setting that should be retrieved in JSP file and that will be reused to initialize the Vue application (See [analytics portlet JSP](https://github.com/Meeds-io/analytics/blob/develop/analytics-webapps/src/main/webapp/WEB-INF/jsp/analytics.jsp) for more details)
### 3.2.8. Available programs Portlet

This will be a simple Vue Portlet that will retrieve enabled programs for current user sorted by creation date through REST API call.

### 3.2.9. Don't miss it Portlet

This will be a simple Vue Portlet that will retrieve enabled challenges for current user through REST API call.

### 3.2.10. Top Challengers Portlet
This new portlet will reuse the same components defined in leaderboard portlet used in stream page. Some UI adaptation will be made to impact this re-usable component, such as currently authenticated user color.

## 3.3. Services & processing
 The portlet preferences has to be managed by GateIN Portal API. Each defined portlet has to define default values for each portlet preference.
No jobs, asynchronous data processing will be needed in the context of current MIP.

## 3.4.1. Macro data model
  A new column `AUDIENCE_ID` will be added in table `GAMIFICATION_DOMAIN` of type `BIGINT` that will reference the Identity Id of the program audience. No foreign key will be used here.
## 3.4.2. Cache management
The DTO of Gamification Domain that will be cached in Storage layer will have to reference the id of identity and not the identity itself to reduce the complexity of domains cache management.

boubaker:


services/src/main/java/org/exoplatform/addons/gamification/storage/dao/GamificationHistoryDAO.java
```diff
@@ -34,11 +34,14 @@
...
 import org.exoplatform.addons.gamification.entities.domain.effective.GamificationActionsHistory;
 import org.exoplatform.addons.gamification.service.dto.configuration.RealizationsFilter;
+import org.exoplatform.addons.gamification.service.dto.configuration.constant.EntityType;
 import org.exoplatform.addons.gamification.service.dto.configuration.constant.HistoryStatus;
 import org.exoplatform.addons.gamification.service.effective.PiechartLeaderboard;
 import org.exoplatform.addons.gamification.service.effective.ProfileReputation;
 import org.exoplatform.addons.gamification.service.effective.StandardLeaderboard;
 import org.exoplatform.commons.persistence.impl.GenericDAOJPAImpl;
+import org.joda.time.DateTimeConstants;
+import org.joda.time.LocalDate;
```
rdenarie: Should we use java.time instead of joda.time ?
As explained here https://github.com/JodaOrg/joda-time, jodatime library was previously used in before java 8.

> Joda-time is no longer in active development except to keep timezone data up to date. From Java SE 8 onwards, users are asked to migrate to java.time (JSR-310) - a core part of the JDK which replaces this project.

services/src/main/resources/locale/addon/Gamification_en.properties
```diff
@@ -31,6 +31,7 @@ exoplatform.gamification.gamificationinformation.Date=Date
 exoplatform.gamification.gamificationinformation.domain.edit.label.description=Enter description
 exoplatform.gamification.gamificationinformation.domain.edit.label.title=Enter domain's title
 exoplatform.gamification.gamificationinformation.Domain=Domain
+exoplatform.gamification.gamificationinformation.domain.Autres=Autres
```
rdenarie: Should be translated in english ?
```suggestion
exoplatform.gamification.gamificationinformation.domain.Autres=Others
```

rdenarie: in addition, the key should also be in english

rdenarie (changes requested):

services/src/main/resources/locale/addon/Gamification_en.properties
boubaker: Fixed, thanks

services/src/main/java/org/exoplatform/addons/gamification/storage/dao/GamificationHistoryDAO.java
boubaker: You are right, I wasn't aware of `LocalDate` that was used from Joda time instead of Java Time. Thanks.
In fact, we have to make the cleanup to delete this from dependencies even.

fixed.

rdenarie (approved):

PR #741, opened 2022-12-23
feat: Adjust Alignment of challenges and programs items  - MEED-1380 - Meeds-io/MIPs#10

ArfaouiMohamedAmine:
Prior to this change, when checking the Alignment of challenges and programs items in overview page, a misalignment is noticed.


boubaker:
Note that description has been rewritten

boubaker (approved):

PR #746, opened 2022-12-26
fix: Unit test fails because of wrong period calculation EXO-61497

ahamdi:
The test **GamificationHistoryDAOTest.testFindMostRealizedRuleIds** failed when run on Sunday because the selected period was wrongly ommitting all realiations done on Sunday. 
The fix changes the period calculation to add the whole day until midnight instead of midnight which is equal to the first second of the day itself (whole day will be ommitted then).


hbenali (approved):
Wow, this is insane IMO :D

PR #761, opened 2023-01-04
fix: Deleted program are displayed on portlet top programs - MEED-1461 - Meeds-io/MIPs#10 

ArfaouiMohamedAmine:
this change is going to fix retrieving deleted domains and this change is going to match the preload to the needed query

PR #767, opened 2023-01-05
fix: Deleted program are displayed on portlet top programs - MEED-1461 - Meeds-io/MIPs#10  (#761)

ArfaouiMohamedAmine:
this change is going to fix retrieving deleted domains and this change is going to match the preload to the needed query


AzmiTouil:
should be backported to develop @ArfaouiMohamedAmine

PR #879, opened 2023-02-01
fix: Engagement Center Challenge activity message not update user's name after profile update - MEED-1186 - Meeds-io/meeds#142

AzmiTouil:
Clear activity cache after updating earner profile

### DIFF
--- a/services/src/main/java/org/exoplatform/addons/gamification/listener/social/profile/GamificationProfileListener.java
+++ b/services/src/main/java/org/exoplatform/addons/gamification/listener/social/profile/GamificationProfileListener.java
@@ -27,117 +27,115 @@ import org.exoplatform.social.core.manager.IdentityManager;
 import org.exoplatform.social.core.profile.ProfileLifeCycleEvent;
 import org.exoplatform.social.core.profile.ProfileListenerPlugin;
 import org.exoplatform.social.core.space.spi.SpaceService;
+import org.exoplatform.social.core.storage.api.ActivityStorage;
 import org.exoplatform.social.core.storage.cache.CachedActivityStorage;
 import java.util.List;
 
 public class GamificationProfileListener extends ProfileListenerPlugin {
 
-    protected RuleService ruleService;
-    protected IdentityManager identityManager;
-    protected SpaceService spaceService;
-    protected GamificationService gamificationService;
-    protected AnnouncementService announcementService;
-    private CachedActivityStorage cachedActivityStorage;
+  protected RuleService         ruleService;
 
+  protected IdentityManager     identityManager;
 
-    public GamificationProfileListener(RuleService ruleService,
-                                       IdentityManager identityManager,
-                                       SpaceService spaceService,
-                                       GamificationService gamificationService,
-                                       AnnouncementService announcementService,
-                                       CachedActivityStorage cachedActivityStorage) {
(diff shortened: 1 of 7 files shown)


services/src/test/java/org/exoplatform/addons/gamification/listener/GamificationProfileListenerTest.java
```diff
@@ -0,0 +1,55 @@
+package org.exoplatform.addons.gamification.listener;
```
boubaker, line 17: Missing license header

services/src/main/java/org/exoplatform/addons/gamification/listener/social/profile/GamificationProfileListener.java
```diff
@@ -18,26 +18,40 @@
...
-
-    public GamificationProfileListener() {
-        this.ruleService = CommonsUtils.getService(RuleService.class);
-        this.identityManager = CommonsUtils.getService(IdentityManager.class);
-        this.spaceService = CommonsUtils.getService(SpaceService.class);
-        this.gamificationService = CommonsUtils.getService(GamificationService.class);
+    protected AnnouncementService announcementService;
+    private CachedActivityStorage cachedActivityStorage;
+
+
```
boubaker: extra line

```suggestion
```

services/src/main/java/org/exoplatform/addons/gamification/listener/social/profile/GamificationProfileListener.java
```diff
@@ -112,5 +127,17 @@ public void aboutMeUpdated(ProfileLifeCycleEvent event) {
...
+      if (CollectionUtils.isEmpty(announcements)) {
+        return;
+      }
+      announcements.forEach(announcement -> clearCache(String.valueOf(announcement.getActivityId())));
+    }
 
+    private void clearCache(String activityId) {
+      if (cachedActivityStorage != null) {
+        cachedActivityStorage.clearActivityCached(activityId);
+      }
```
boubaker: if `CachedActivityStorage` is added in constructor as dependency, then it can't be null. Besides, it's may be better to add this methods in Service Layer in ActivityManager API.

BTW, it was already treated in Kudos here: https://github.com/Meeds-io/kudos/pull/287/files if this helps as well to use the same behavior.

```suggestion
      cachedActivityStorage.clearActivityCached(activityId);
```

services/src/test/java/org/exoplatform/addons/gamification/listener/GamificationProfileListenerTest.java
AzmiTouil: Done!

services/src/main/java/org/exoplatform/addons/gamification/listener/social/profile/GamificationProfileListener.java
AzmiTouil: Done!

services/src/main/java/org/exoplatform/addons/gamification/listener/social/profile/GamificationProfileListener.java
AzmiTouil: Updated

boubaker (approved):

PR #882, opened 2023-02-03
fix: Engagement Center Default rules are not available anymore - MEED-1663

AzmiTouil:
Prior to this change, there was a problem when creating the default event (automatic rule), it is due to the suppression of defaults programs
+ Ensure rules and badges refer to the program ID rather than the program name.
+ Update realization display in case of the action or the program name has been changed


services/src/main/java/org/exoplatform/addons/gamification/service/setting/rule/impl/RuleRegistryImpl.java
```diff
@@ -109,15 +109,7 @@ private void store(RuleConfig ruleConfig, RuleDTO ruleDTO) {
           ruleDto.setEvent(ruleConfig.getEvent());
           ruleDto.setLastModifiedBy("Gamification");
           ruleDto.setCreatedBy("Gamification");
-          ruleDto.setArea(ruleConfig.getZone());
```
boubaker, line 112: we may need to this in the future, should we test on null value of zone only instead of deleting the init behavior?

AzmiTouil: Done!

services/src/main/java/org/exoplatform/addons/gamification/service/setting/rule/impl/RuleRegistryImpl.java
```diff
@@ -109,15 +109,10 @@ private void store(RuleConfig ruleConfig, RuleDTO ruleDTO) {
...
-          ruleDto.setEnabled(true);
```
boubaker, line 113: need to check if `enabled = false` should be the default value of imported rules, is it ?

services/src/main/java/org/exoplatform/addons/gamification/service/configuration/RuleServiceImpl.java
```diff
@@ -190,13 +190,11 @@ public RuleDTO createRule(RuleDTO ruleDTO, String username) throws IllegalAccess
     }
     ruleDTO.setCreatedBy(username);
     ruleDTO.setLastModifiedBy(username);
-    DomainDTO domainDTO = Utils.getDomainByTitle(ruleDTO.getArea());
```
boubaker, line 193: what about existing installations, should there be an upgrade plugin to change `Area` fiels of existing rules to use `programId` ? If yes, we should delete completely this field from database & DTO if the Model has evolved to keep the id only as unique way to link a rule to a program + we should add a foreign key on it for better evolutivity, coherence and maintainability.

services/src/main/java/org/exoplatform/addons/gamification/service/setting/rule/impl/RuleRegistryImpl.java
AzmiTouil: It's a duplication, already added https://github.com/Meeds-io/gamification/pull/882/files#diff-d621bebee8d9528d485209b43ecfb65ceb0a7cb605380415b7f360eca1e4ad46R108

services/src/main/java/org/exoplatform/addons/gamification/service/configuration/RuleServiceImpl.java
AzmiTouil: Done!

boubaker (approved):

PR #956, opened 2023-03-22
fix: Migration from 6.3 to 6.4,user does not have permission to announce a challenge - EXO-62903 - Meeds-io/meeds#846

IlhemEssaadi:
Prior to this change, when the user announced a challenge with version 6.3.5 and then migrated to version 6.4, he was not allowed to announce the same challenge because in version 6.3 it is the challenge that has an audience and not the program.After this change, the announcement was successfully published.

### DIFF
--- a/packaging/pom.xml
+++ b/packaging/pom.xml
@@ -22,7 +22,7 @@ Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
     <parent>
         <groupId>org.exoplatform.addons.gamification</groupId>
         <artifactId>gamification</artifactId>
-        <version>2.5.x-exo-SNAPSHOT</version>
+        <version>2.5.x-maintenance-SNAPSHOT</version>
     </parent>
 
     <artifactId>gamification-packaging</artifactId>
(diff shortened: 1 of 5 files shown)


boubaker:
AFAIK, this is expected to not being able to announce a challenge not being part of a program with an audience anymore. @srenault-meeds correct me please if I'm wrong.

srenault-meeds:
Yes indeed @IlhemEssaadi 
This has to be done by interface by functional consultant or by reward admin.
cc @margondicco please can you make sure exo team doesn't change business rules regarding gamification without telling us so we coordinate ourselves the solution?

Jihed525:
@srenault-meeds this PR was treated as reported in your exo-feedback https://community.exoplatform.com/portal/dw/tasks/taskDetail/62740 for migration cases and it was considered as High Priority, it is no longer needed ?
I might have missed a detail @margondicco @boubaker ?

rdenarie:
Hello
@Jihed525 In fact, there is a real problem after 6.4.0 migration
But, the provided fix does not follow the correct way because it seems that it revert a wanted feature.

I start a deeper analysis on customer environment which where migrated to try to find the exact problem.

PR #1011, opened 2023-05-16
fix: Fix accessing program detail when not program member - MEED-2158 - Meeds-io/MIPs#49

boubaker:
Prior to this change, a simple user can directly access a program through direct URL. This change will ensure to secure program detail screen access by checking user permissions in REST call.


boubaker:
Waiting for Task & Meeds issue creation.

PR #1045, opened 2023-05-30
fix: Capital letters aren't considered for deutsch language in challenge tab - EXO-64389 - Meeds-io/meeds#951

akhanfir:
Prior to this change, when set plf language to german/Deutsch, go to challenge app and select the challenges tab 'herausforderungen' then check the button label, The button label is 'Fügen Sie eine herausforderung hinzu' which does not respect the capital letters rule in German language. To fix this problem, when plf language is german/Deutsch add a class text-none to eliminate existing text transformations knowing that this change keeps uppercase applied to the first letter of the text . After this change, the text is well written with capital letters in crowdin yet it is changed by css .

### DIFF
--- a/portlets/src/main/webapp/vue-app/engagement-center/components/challenges/Challenges.vue
+++ b/portlets/src/main/webapp/vue-app/engagement-center/components/challenges/Challenges.vue
@@ -28,7 +28,9 @@ Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
           class="btn btn-primary"
           @click="openChallengeDrawer">
           <v-icon small>fas fa-plus</v-icon>
-          <span class="mx-2 d-none d-lg-inline text-capitalize-first-letter subtitle-1">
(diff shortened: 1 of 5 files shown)


portlets/src/main/webapp/vue-app/engagement-center/components/challenges/Challenges.vue
```diff
@@ -28,7 +28,9 @@ Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
           class="btn btn-primary"
           @click="openChallengeDrawer">
           <v-icon small>fas fa-plus</v-icon>
-          <span class="mx-2 d-none d-lg-inline text-capitalize-first-letter subtitle-1">
```
ahamdi, line 31: shouldn't we just remove the class **text-capitalize-first-letter** and ask to change the text in Crowdin ?

hbenali:
Stale PR

PR #1229, opened 2023-10-02
fix: Fix Action Form Disable Button - MEED-2596 - Meeds-io/meeds#1135

boubaker:
Prior to this change, the action form button was disabled after changing the translation of name or description. In fact, the modifications made in translation drawer wasn't detected as a modification. This change will include translations in the comparaison algorithm to allow update the action.

### DIFF
--- a/portlets/src/main/webapp/vue-app/rules/components/drawers/RuleFormDrawer.vue
+++ b/portlets/src/main/webapp/vue-app/rules/components/drawers/RuleFormDrawer.vue
@@ -499,7 +499,6 @@ export default {
       }
     },
     ruleDescription() {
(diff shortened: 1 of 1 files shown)


portlets/src/main/webapp/vue-app/rules/components/drawers/RuleFormDrawer.vue
```diff
@@ -465,6 +499,7 @@ export default {
       }
     },
     ruleDescription() {
+      console.warn('ruleDescription', this.ruleDescription);
```
boubaker:
```suggestion
```

MayTekayaa (approved):

PR #1355, opened 2023-11-27
fix: Update the applications images in Contributions category of page editor - EXO-67096

IlhemEssaadi:
After this change, we update the default images for each application in the Contributions category.


azayati (approved):
LGTM

PR #1384, opened 2023-12-08
feat: Enhance Achievements filter UX - MEED-3051 - Meeds-io/meeds#139…

rdenarie:
…6 (#1381)

This change will allow to filter by rules/actions in achivements list. In addition, this will allow to select filters while browsing achievements.
In addition, the ES index is used to search full text in rules in order to retrieve the rule ids and then apply filtered ids on a database query to centralize the query filter processing. Thus, we may have deleted or disabled rules in ES to allow retrieving it while searching using Achievements UI.


azayati (approved):
LGTM

PR #1506, opened 2024-03-13
Merge meed 3423

AzmiTouil:


services/src/main/java/io/meeds/gamification/entity/RealizationEntity.java
```diff
@@ -58,7 +58,7 @@
       + " SELECT earnerId, ROW_NUMBER() OVER() rankIndex FROM " 
       + " (SELECT g.EARNER_ID earnerId, SUM(g.ACTION_SCORE) total"
       + " FROM GAMIFICATION_ACTIONS_HISTORY g"
-      + " WHERE g.CREATED_DATE >= :date"
+      + " WHERE (g.CREATED_DATE >= :date OR g.LAST_MODIFIED_DATE >= :date)"
```
boubaker: It would be better to not multiply the fields used in this query knowing that it's too much expensive already and highly used. Let's discuss more about further options

AzmiTouil: Updated!

boubaker (dismissed):

boubaker (approved):

PR #1577, opened 2024-05-11
fix: Fix Realization Object Id Analytics Mapping Type - MEED-6810 - Meeds-io/meeds#1970

boubaker:
Prior to this change, the contributionObjectId field uses mainly ids of type Long, whereas, some Gamification connectors may need to use String type for this field to add more details. This change forces analytics data type of this field to be mapped as keyword instead.


Christyempire:
When mapping how do I realised the objects assigned

PR #1801, opened 2025-02-07
feat: add username retrieval from connector user identifier REST API method

hbenali:
This PR introduces a new REST API method to retrieve a username based on a given connector user identifier to allow external integrations.

### DIFF
--- a/services/src/main/java/io/meeds/gamification/rest/ConnectorRest.java
+++ b/services/src/main/java/io/meeds/gamification/rest/ConnectorRest.java
@@ -94,6 +94,35 @@ public class ConnectorRest implements ResourceContainer {
   }
 
   @GET
+  @Produces(MediaType.TEXT_PLAIN)
+  @RolesAllowed("users")
+  @Path("username/{connectorName}")
+  @Operation(summary = "Retrieve the username associated with a connector user identifier.", description = "Fetches the username corresponding to the given connector user identifier.", method = "GET")
+  @ApiResponses(value = {
+      @ApiResponse(responseCode = "200", description = "Request fulfilled"),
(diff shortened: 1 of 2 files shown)


AzmiTouil (approved):

boubaker:
We should have made this a personal choice, whether the user wants to expose or not the connected accounts names, like the preference in github to expose the email or not, or like any profile setting attribute hide/show preference already existing in the platform. This can make have a privacy concern.

hbenali:
> We should have made this a personal choice, whether the user wants to expose or not the connected accounts names, like the preference in github to expose the email or not, or like any profile setting attribute hide/show preference already existing in the platform. This can make have a privacy concern.

I understand your concern. When a user adds their account, I believe they intentionally approve the sharing of their access, especially since this type of information is displayed publicly on their profile. So, they are aware of granting access to others. In my opinion, should we consider adding more restrictions, like limiting access to this method to admins only? WDYT?

boubaker:
> > We should have made this a personal choice, whether the user wants to expose or not the connected accounts names, like the preference in github to expose the email or not, or like any profile setting attribute hide/show preference already existing in the platform. This can make have a privacy concern.
> 
> I understand your concern. When a user adds their account, I believe they intentionally approve the sharing of their access, especially since this type of information is displayed publicly on their profile. So, they are aware of granting access to others. In my opinion, should we consider adding more restrictions, like limiting access to this method to admins only? WDYT?

Yes please, this can decrease considerably the criticality :+1:

boubaker:
> Yes please, this can decrease considerably the criticality 👍

Especially to not allow easily iterate through user accounts to retrieve their respective accounts on other third party systems (allowing this for admins only, could be sufficient for now)

PR #1940, opened 2025-08-13
fix: Fix the display of video youtube in a quest - MEED-9480 - Meeds-io/meeds#3501

MayTekayaa:
This change will fix the display of video youtube in a quest.

### DIFF
--- a/services/src/main/java/io/meeds/gamification/rest/builder/RuleBuilder.java
+++ b/services/src/main/java/io/meeds/gamification/rest/builder/RuleBuilder.java
@@ -155,10 +155,9 @@ public class RuleBuilder {
                                                                            Utils.getCurrentUser());
     translatedLabels(translationService, rule, locale);
 
-    String description = rule.getDescription();
     return new RuleRestEntity(rule.getId(),
                               rule.getTitle(),
-                              processRichEditorContent(xmlProcessor, description),
+                              rule.getDescription(),
                               rule.getScore(),
                               program,
                               rule.isEnabled(),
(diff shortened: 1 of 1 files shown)


services/src/main/java/io/meeds/gamification/rest/builder/RuleBuilder.java
```diff
@@ -220,10 +219,6 @@ public static UserInfoContext toUserContext(ProgramService programService,
     return userContext;
   }
 
-  public static String processRichEditorContent(XMLProcessor xmlProcessor, String content) {
```
boubaker, line 223: I think we may have a regression on Email Notification content, can you double check please. In addition, can you check please if an XSS attack is possible when removing this.
This change was introduced using the commit https://github.com/Meeds-io/gamification/commit/3184ce8c2de08792c642ef2b815241f201b67a2c

As described:
> This change will transform Action descriptions before retrieving to UI to auto determine the links targets.

Can you ensure that the links targets are ok switch the needed behavior (see https://builders.meeds.io/portal/meeds/tasks/taskDetail/2264)

MayTekayaa: I have checked links targets and the behavior is ok, i will also check for email notification.

MayTekayaa: All the scenarios I did were good for me (same behavior on ACC) @boubaker

boubaker (approved):

PR #1984, opened 2026-07-06
feat: Add Gamification MCP tools for EVA (campaigns & quests) - EXO-88395

bmestrallet:
## What

14 MCP tools in a new `io.meeds.gamification.mcp` package (`gamification/services`) exposing the Gamification add-on to EVA, using the product's **UI vocabulary**: **campaign** = `Program`, **quest** = `Rule`. Each is a Spring `@Service @Profile("mcp-server")` implementing `McpToolPlugin`, acting **as the current user**; writes are approval-gated and ACL-enforced.

**Reads (readOnlyHint):** `list_campaigns`, `get_campaign`, `list_quests`, `get_quest`, `get_my_gamification_score`, `get_gamification_leaderboard`, `list_my_realizations`, `list_my_announcements`.

**Writes (require_approval):** `create_campaign` (→ `canAddProgram`), `create_quest`, `update_quest` (fetch-mutate-save), `delete_quest` (→ program-owner/`canEditRule`), `announce_quest`, `cancel_quest_announcement`.

- Quests are **MANUAL-only** for now; AUTOMATIC (event-triggered) creation is rejected with a clear message.
- **Hard guard:** the no-ACL `RealizationService.createRealizations` self-award path is never exposed (a user must not be able to grant themselves points).

## Wiring
- `services/pom.xml`: `mcp-server-tools` (provided) + the **`-parameters`** compiler flag.
- Parent `pom.xml`: `mcp-server` dependencyManagement import. Branch already on `7.3.x-ai-contribution-SNAPSHOT`.

## Tests & verification
- `mvn clean install -pl services -Pcoverage` → BUILD SUCCESS, 333 tests / 0 failures, coverage gate **0.75** held; `-parameters` confirmed (`javap` MethodParameters > 0).
- A **kernel/DB integration test** asserts `create_campaign` → `get_campaign`/`list_campaigns` read-back works immediately (no cache gap).
- **Verified live with EVA:** creating a "Q3 Contributor" campaign + 3 quests persisted and rendered in the Contribution Campaigns portlet.

AI contribution. Branch off `feature/ai-contribution`.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -53,6 +53,7 @@ Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
         <io.meeds.social.version>7.3.x-ai-contribution-SNAPSHOT</io.meeds.social.version>
         <io.meeds.platform-ui.version>7.3.x-ai-contribution-SNAPSHOT</io.meeds.platform-ui.version>
         <addon.meeds.analytics.version>7.3.x-ai-contribution-SNAPSHOT</addon.meeds.analytics.version>
+        <addon.meeds.mcp-server.version>7.3.x-ai-contribution-SNAPSHOT</addon.meeds.mcp-server.version>
 
         <!-- Sonar properties -->
         <sonar.organization>meeds-io</sonar.organization>
@@ -66,6 +67,15 @@ Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
               <groupId>io.meeds.social</groupId>
               <artifactId>social</artifactId>
               <version>${io.meeds.social.version}</version>
+              <type>pom</type>
+              <scope>import</scope>
+            </dependency>
+
+            <!-- MCP Server API -->
+            <dependency>
+              <groupId>io.meeds.mcp-server</groupId>
+              <artifactId>mcp-server</artifactId>
+              <version>${addon.meeds.mcp-server.version}</version>
               <type>pom</type>
               <scope>import</scope>
             </dependency>
--- a/services/pom.xml
+++ b/services/pom.xml
@@ -71,6 +71,13 @@
       <artifactId>analytics-api</artifactId>
       <scope>provided</scope>
     </dependency>
+
+    <!-- MCP Server API -->
+    <dependency>
+      <groupId>io.meeds.mcp-server</groupId>
+      <artifactId>mcp-server-tools</artifactId>
+      <scope>provided</scope>
+    </dependency>
     <dependency>
       <groupId>io.meeds.analytics</groupId>
       <artifactId>analytics-listeners</artifactId>
@@ -101,6 +108,15 @@
   <build>
     <finalName>${project.artifactId}</finalName>
     <plugins>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-compiler-plugin</artifactId>
+        <configuration>
+          <compilerArgs>
+            <arg>-parameters</arg>
+          </compilerArgs>
+        </configuration>
+      </plugin>
       <plugin>
         <groupId>io.openapitools.swagger</groupId>
         <artifactId>swagger-maven-plugin</artifactId>
--- a/services/src/main/java/io/meeds/gamification/mcp/GamificationMcpTool.java
+++ b/services/src/main/java/io/meeds/gamification/mcp/GamificationMcpTool.java
@@ -111,7 +111,13 @@ public class GamificationMcpTool implements McpToolPlugin {
                                            Long spaceId,
                                            Integer limit,
                                            Integer offset) throws IllegalAccessException {
-    ProgramFilter filter = new ProgramFilter(spaceId == null);
+    // Mirror ProgramRest#getPrograms: always build the filter with
+    // allSpaces=false (the no-arg ProgramFilter default) so the DAO keeps the
+    // "(audienceId IS NULL OR visibility = OPEN)" visibility restriction and
+    // the service scopes to the caller's member spaces. Passing allSpaces=true
+    // (the previous spaceId==null default) would leak RESTRICTED programs from
+    // spaces the caller is not a member of.
+    ProgramFilter filter = new ProgramFilter();
     filter.setStatus(EntityStatusType.ALL);
     if (StringUtils.isNotBlank(term)) {
       filter.setProgramTitle(term.trim());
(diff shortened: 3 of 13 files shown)


boubaker:
## Review: Gamification MCP tools for EVA

### Overview
Adds 14 MCP tools (`io.meeds.gamification.mcp.GamificationMcpTool`, `@Service @Profile("mcp-server")`) exposing campaigns (Programs) and quests (Rules) to the EVA agent, plus DTOs, `ai-tool-definitions.json`, a Mockito unit test suite, and a kernel/DB integration test. Wiring (`pom.xml`, `services/pom.xml`, `-parameters` compiler flag) is clean and matches the `mcp-server-tools` integration pattern.

---

### 🔴 Critical — ACL bypass in `listCampaigns` / `listQuests`

`GamificationMcpTool.java:114` and `:146`:
```java
ProgramFilter filter = new ProgramFilter(spaceId == null);   // listCampaigns
RuleFilter filter = new RuleFilter(campaignId == null);      // listQuests
```
Both constructors set the filter's `allSpaces` field. Tracing how that flows through `ProgramServiceImpl.computeUserSpaces` / `RuleServiceImpl.computeUserSpaces`:
- `allSpaces` is only explicitly forced for rewarding-manager users; for ordinary users it is **never reset**, only `spacesIds`/`spaceIds` gets narrowed to the caller's own memberships.
- `ProgramDAO.buildPredicates` / `RuleDAO.buildPredicates` only add the `(audienceId IS NULL OR visibility = OPEN)` restriction `else if (!filter.isAllSpaces())` — when `allSpaces=true`, **no visibility predicate is added at all**.

Net effect: any call to `list_campaigns` or `list_quests` **without** an explicit `space_id`/`campaign_id` — the normal, default way an AI agent would ask "what campaigns/quests are there?" — sets `allSpaces=true` and causes the query to return `RESTRICTED` programs/rules from spaces the current user is **not** a member of. Every genuine REST endpoint (`ProgramRest`, `RuleRest`) constructs its filter with the default `allSpaces=false`; this is the only end-user-facing code path in the codebase that sets it to `true`.

This directly contradicts the PR description's "ACL-enforced" claim and is a real information-disclosure bug — a user could ask EVA to list campaigns/quests and get titles/descriptions/budgets from spaces they can't otherwise see.

**Fix**: build these filters the same way the REST layer does (default `allSpaces=false`), and only set `spacesIds`/`programId` when an id is explicitly provided — don't use "no id given" as a proxy for "search everywhere".

---

### 🟠 High — the "core regression" test never runs

`GamificationMcpToolKernelTest` (the container/DB-backed test that's supposed to prove the read-after-write/cache-invalidation fix — literally documented in its own Javadoc as "The core regression") extends `AbstractServiceTest` but is **not** added to `InitContainerTestSuite`'s `@SuiteClasses` list (only the plain Mockito `GamificationMcpToolTest` was added).

In this module, `maven-surefire-plugin` is configured with `<includes><include>**/InitContainerTestSuite.java</include></includes>` — no `**/*Test.java` wildcard — and every other container-backed test class in the module (60+) is enumerated in that one suite. Since `GamificationMcpToolKernelTest` isn't listed anywhere, `mvn test`/`mvn verify` will **never execute it** — it's dead code. The PR's verification claim ("kernel/DB integration test asserts... no cache gap") is not actually backed by CI.

**Fix**: add `GamificationMcpToolKernelTest.class` to `InitContainerTestSuite`'s `@SuiteClasses`.

---

### 🟡 Medium — inconsistent ACL-exception handling across tools

Several methods declare `throws IllegalAccessException` but internally catch every possible source of it and rethrow `IllegalStateException` instead, so the declared checked exception can never actually propagate:
- `getCampaign` (via `resolveProgram`)
- `getQuest` (via `resolveRule`)
- `createCampaign`, `createQuest`, `updateQuest`, `announceQuest`

Meanwhile `listCampaigns`, `getGamificationLeaderboard`, `listMyRealizations`, `listMyAnnouncements` let `IllegalAccessException` propagate untouched from the underlying service call. If the MCP runtime maps `IllegalAccessException` to a specific "permission denied" tool-error type (mirroring the REST 403 contract), this inconsistency means half the write/get tools will report ACL denial as a generic error instead. Worth checking against `McpToolPlugin`'s actual exception contract, and converting all tools the same way.

Also, `createQuest`'s `canManageCampaign` proactive check duplicates the reactive `catch (IllegalAccessException e)` around `ruleService.createRule` — harmless but redundant; ACL enforcement should live in the Service layer, not be re-implemented in the tool layer.

---

### Minor / style
- `ai-tool-definitions.json` and the Javadoc are clear and well-written; tool descriptions correctly point the agent to the right "resolve with..." tool for each id.
- DTOs are simple and appropriately thin (ids-only for nested references), consistent with the Storage-DTO simplicity convention.
- Good defensive handling of boxed params throughout (`clampLimit`, `clampOffset`, null-safe parsing).
- The "Hard guard" claim (never exposing `RealizationService.createRealizations`'s self-award path) checks out — no tool method calls it.

---

### Test coverage
- The Mockito unit tests (`GamificationMcpToolTest`) are thorough — good coverage of validation, ACL-denial-translation, and not-found paths for every tool.
- As noted above, the one test designed to catch the actual regression this PR's Storage-layer caching depends on is not wired into the build.
- No test exercises the `allSpaces`/visibility scoping described above — worth adding a kernel test asserting that `list_campaigns`/`list_quests` do **not** return a `RESTRICTED` program/rule from a space the caller isn't a member of.

---

**Bottom line**: solid, well-tested shape overall, but the `allSpaces` ACL bypass is a real security bug that should block merge, and the dead kernel test should be wired in before relying on it as regression coverage.

---
🤖 This review was generated by [Claude Code](https://claude.com/claude-code).

bmestrallet:
@boubaker — thanks, both fixed (commit 416756ac):

- **🔴 Critical — ACL bypass:** `listCampaigns`/`listQuests` now build `new ProgramFilter()` / `new RuleFilter()` (no-arg → default `allSpaces=false`), exactly as `ProgramRest#getPrograms` / `RuleRest#getRules` do. Verified against `ProgramDAO`/`RuleDAO.buildPredicates` + `computeUserSpaces` that the default still returns the caller’s accessible set (member spaces + OPEN/no-audience) — not an empty list — and no longer leaks RESTRICTED items from non-member spaces.
- **🟠 Dead kernel test:** `GamificationMcpToolKernelTest` is now wired into `InitContainerTestSuite`, so CI actually runs it.
- **Regression coverage:** added captor tests asserting the filter is built with `isAllSpaces()==false`, plus a DB-backed `testListDoesNotLeakRestrictedContentToNonMember` (RESTRICTED program+quest, non-member session, asserts neither leaks). Both fail against the pre-fix code. Full suite 338 green.

On the 🟡 Medium exception consistency: I left the write tools as-is for now — existing tests explicitly assert the `IllegalStateException` behavior, so flipping to propagate `IllegalAccessException` would be a behavior change I didn’t want to bundle with the security fix. Happy to do it as a separate follow-up if you want the REST-style contract.

